PR #38, opened 2022-09-23
Fix: Correct driver name

Huyston:


jbms:
Thanks for catching this.  The gh-pages branch is auto-generated from the restructured text sources in the master branch, though, so any changes have to be made there.

Huyston:
My bad. Corrected the source now. Closing this because I made another PR.

PR #62, opened 2022-11-07
Initial OmeTiff support

sameeul:
Hello, I added the basic ometiff read support by creating a new `ometiff` driver and a new `tiled_tiff` kvstore. The key difference between this and the existing `tiff` image driver is the tiled access which allows to read partial tiles of a large file instead of reading the whole file in memory. I mimicked some unit tests from n5. There are few n5 test cases I could not make to pass. I would like to get feedback on the PR and eventually get it merged.


third_party/com_github_pugixml/workspace.bzl
```diff
@@ -0,0 +1,19 @@
+load("//third_party:repo.bzl", "third_party_http_archive")
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
+
+def repo():
+    maybe(
+        third_party_http_archive,
+        name = "com_github_pugixml",
+        urls = [
```
laramiel, line 8: This needs to add strip_prefix, and then remove the prefix manipulation from bundled.BUILD.bazel.

   `strip_prefix = "pugixml-1.13",`

third_party/com_github_pugixml/bundled.BUILD.bazel
```diff
@@ -0,0 +1,11 @@
+licenses(["notice"])
```
laramiel, line 1: I'd prefer that this was named `pugixml.BUILD.bazel`; I think that's more discoverable than `bundled.BUILD.bazel`, though I haven't renamed the older BUILD files.

third_party/third_party.bzl
```diff
@@ -35,6 +35,7 @@ load("//third_party:pypa/workspace.bzl", repo_pypa = "repo")
 load("//third_party:rules_perl/workspace.bzl", repo_rules_perl = "repo")
 load("//third_party:rules_pkg/workspace.bzl", repo_rules_pkg = "repo")
 load("//third_party:se_curl/workspace.bzl", repo_se_curl = "repo")
+load("//third_party:com_github_pugixml/workspace.bzl", repo_com_github_pugixml = "repo")
```
laramiel, line 38: Sort these

tensorstore/driver/ometiff/metadata.h
```diff
@@ -0,0 +1,134 @@
...
+#include "tensorstore/data_type.h"
+#include "tensorstore/internal/json_binding/bindable.h"
+#include "tensorstore/schema.h"
+#include "tensorstore/serialization/fwd.h"
+#include "tensorstore/strided_layout.h"
+#include "tensorstore/util/garbage_collection/fwd.h"
+#include "tensorstore/util/result.h"
+#include "tensorstore/util/span.h"
+
+namespace tensorstore {
```
laramiel, line 21: All your utilities should live in this namespace.

tensorstore/kvstore/tiled_tiff/tiled_tiff_key_value_store.cc
```diff
@@ -0,0 +1,389 @@
+#include <tiffio.h>
```
laramiel, line 1: This is a good starting point. 

Generally you will want to make it work with an underlying kvstore, so that means it will need a read/write cache. Look at tensorstore/kvstore/neuroglancer_uint64_sharded/neuroglancer_uint64_sharded.cc as a model.

I'd probably start with List, and add metadata manipulation to internal/image/tiff_{encoder,decoder}

third_party/com_github_pugixml/workspace.bzl
sameeul: updated

third_party/com_github_pugixml/bundled.BUILD.bazel
sameeul: updated

third_party/third_party.bzl
sameeul: updated

tensorstore/driver/ometiff/metadata.h
sameeul: moved utilities in omexml.* under `tensorstore::omexml` namespace

tensorstore/kvstore/tiled_tiff/tiled_tiff_key_value_store.cc
sameeul: I am a little confused here. The usecase that motivated us to have a OmeTiff driver is to enable chuck based access. Usually these tiff files are big and at a given time, the user is interested in a small chunk of data. The decoder in internal/image/tiff_decoder seems to be reading the whole file while we want to read only a portion of the file. So, should not we have a separate driver and kvstore? According to the OME spec, these files are always 5D from an API point of view, however, each 2D slices are stored in a separate IFD inside the tiff file. The OmeXml metadata helps us to figure out the IFD value for a given Z, Channel and Time value. And then we can extract the (x_range, y_range) values using the ReadTileFromFile call.

tensorstore/driver/ometiff/BUILD
```diff
@@ -0,0 +1,134 @@
...
+        "//tensorstore/util:span",
+        "//tensorstore/util/garbage_collection",
+        "@com_github_nlohmann_json//:nlohmann_json",
+        "@com_google_absl//absl/algorithm:container",
+        "@com_google_absl//absl/base:endian",
+        "@com_google_absl//absl/status",
+        "@com_google_absl//absl/strings",
+    ],
+    alwayslink = 1,
+    linkopts = ["/usr/lib/x86_64-linux-gnu//libtiffxx.so.5",
```
jbms, line 63: This assumes a system-installed libtiff in a particular location, which we can't do for tensorstore --- instead, just add `@libtiff//:tiff` to `deps` since we already have libtiff available as a dependency.

tensorstore/kvstore/tiled_tiff/tiled_tiff_key_value_store.cc
jbms: It appears that this ometiff driver just accesses the local filesystem directly, which means it would only work if the ometiff dataset is stored on the local filesystem --- instead it should use a base kvstore so that it works with any storage system.

The tensorstore kvstore interface allows you to specify a byte range for read requests to avoid reading the entire file.

sameeul: I also agree that the `ometiff` driver is tightly coupled with tiled `tiled_tiff` kvstore. Here is my difficulty in uncoupling them. 

I want to read a tile as a chunk and that makes me dependent on high-level API like `TiffReadTile(...)`. Also, when I am extracting the OmeXML metadata, I am using `TiffGetField` to read the IMAGE_DESCRIPTION tag. If I use a base kvstore, how should I approach it so that I do not have to borrow the implementation of these functions?

tensorstore/driver/ometiff/BUILD
sameeul: My bad, I actually don't need it here anymore.

tensorstore/kvstore/tiled_tiff/tiled_tiff_key_value_store.cc
sameeul: > This is a good starting point.
> 
> Generally you will want to make it work with an underlying kvstore, so that means it will need a read/write cache. Look at tensorstore/kvstore/neuroglancer_uint64_sharded/neuroglancer_uint64_sharded.cc as a model.
> 
> I'd probably start with List, and add metadata manipulation to internal/image/tiff_{encoder,decoder}

So, if I understand it correctly, `ometiff` driver will still work with a single specific key-value store driver but the users will not be able to choose it via JSON spec, rather it will be sort of hard-coded in `ometiff` similar to `neuroglancer_uint64_sharded`?

jbms: The use of tiled_tiff as an implementation detail is fine --- that should not be exposed to users.  But tiled_tiff should use a base kvstore (that is exposed to users) rather than just using the local filesystem.

sameeul: Got it. Thanks. I am looking at the neuroglacner driver to see that is done and will try to mimic similar design.

sameeul: Hello again, I spent some time to understand the neuroglancer_uint64_unsharded and did not actually get too far. 
Here are some of my questions related to `'ometiff' driver. These may appear very naive questions but I would really appreciate your guidance on them:

1. If the user is allowed to use 'file', 'memory', 'gcs' and 'http' as  key-value store driver and I want to still leverage `libtiff` API to extract metadata and various information from the file and also to read tiles, will it be possible use the `tiled_tiff` key-value storage driver (with some rearrangement)? Or will I need to reimplement most of the `libtiff` functionality here. I am not sure how to how to successfully use the byte-range read functionality of tensorstore interface without bringing in those `libtiff` implementation details (since they actually figures out which part of the file to read)

2. I see that `ShardedKeyValueStore` inherits from `internal_kvstore` and in the neuroglancer tensorstore driver, `GetDataKeyValueStore` is overrided. Is that the path you are suggesting me to take when you mean that `tiled_tiff` should use a base kvstore that is exposed to the users?

jbms: > Hello again, I spent some time to understand the neuroglancer_uint64_unsharded and did not actually get too far. 
> Here are some of my questions related to `'ometiff' driver. These may appear very naive questions but I would really appreciate your guidance on them:
> 
> 1. If the user is allowed to use 'file', 'memory', 'gcs' and 'http' as  key-value store driver and I want to still leverage `libtiff` API to extract metadata and various information from the file and also to read tiles, will it be possible use the `tiled_tiff` key-value storage driver (with some rearrangement)? Or will I need to reimplement most of the `libtiff` functionality here. I am not sure how to how to successfully use the byte-range read functionality of tensorstore interface without bringing in those `libtiff` implementation details (since they actually figures out which part of the file to read)

Libtiff has its own file i/o abstraction, so you could potentially implement the interface libtiff expects on top of a kvstore, support to how the existing non-tiled tiff driver implements it on top of absl::Cord.  The difficulty is that libtiff uses a synchronous API, so that would only allow a single concurrent read operation at a time.  Working around that limitation might require either using some libtiff internals directly, or somehow cloning libtiff handles.

> 
> 2. I see that `ShardedKeyValueStore` inherits from `internal_kvstore` and in the neuroglancer tensorstore driver, `GetDataKeyValueStore` is overrided. Is that the path you are suggesting me to take when you mean that `tiled_tiff` should use a base kvstore that is exposed to the users?

Yes, the use of a base kvstore exposed to users should be similar to how it is done with ShardedKeyValueStore.  But note that GetDataKeyValueStore is part of the kvs_backed_chunk_driver abstraction, which I don't think is a good fit for tiled_tiff.  Instead tiled_tiff should probably use ChunkCache directly.

PR #82, opened 2023-02-28
Replace deprecated `set-output` command

cibinmathew:
**Issue**

Github deprecating `save-state` and `set-output` commands in Github actions by [end of May 2023](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

**Fix**

Convert deprecated save-state and set-output commands as per [workflow-commands-for-github-actions](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,7 @@ jobs:
     - name: Get pip cache dir
       id: pip-cache
       run: |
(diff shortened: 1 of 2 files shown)


laramiel:
This was resolved elsewhere. Thanks.

PR #98, opened 2023-05-08
Add BAZEL_PATH environment variable

carlthome:
Trying to make `tensorstore` work without downloading Bazel through Bazelisk, for sandboxed builds without a writeable $HOME directory.

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -457,6 +457,11 @@ def execute_bazel(bazel_path, argv):
 
 
 def get_bazel_path():
+    
+    # If Bazel already exists, skip Bazelisk wrapper.
+    if "BAZEL_PATH" in os.environ:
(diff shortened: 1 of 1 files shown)


jbms:
For the Python install, there are two existing environment variables you could use to potentially solve this issue:

https://google.github.io/tensorstore/installation.html#envvar-TENSORSTORE_BAZELISK

You could set this to a stub Python script that just calls the desired bazel executable.

https://google.github.io/tensorstore/installation.html#envvar-BAZELISK_HOME

You can set this to change the basket bazelisk cache directory.

You will presumably also need to change the bazel cache directory, using 

https://google.github.io/tensorstore/installation.html#envvar-TENSORSTORE_BAZEL_STARTUP_OPTIONS

carlthome:
Thanks @jbms! Good suggestions! 🎉

PR #105, opened 2023-07-07
 Update TensorStore 20xy copyright notice year to 2023

sjperkins:
xref https://github.com/google/tensorstore/pull/91#discussion_r1248141243

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright 2022 The TensorStore Authors
+# Copyright 2023 The TensorStore Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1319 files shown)


jbms:
Sorry for the confusion --- the legal guidance from our corporate counsel as far as I know, is that copyright years don't need to be updated in source files even if they are modified in a later year. For this project, therefore, we just list the copyright year in each file as the first year in which the copyright applies, i.e. the current year at the time the file is created.

sjperkins:
> Sorry for the confusion --- the legal guidance from our corporate counsel as far as I know, is that copyright years don't need to be updated in source files even if they are modified in a later year. For this project, therefore, we just list the copyright year in each file as the first year in which the copyright applies, i.e. the current year at the time the file is created.

OK, thanks :-)

Issue #114, opened 2023-09-21
Further S3 Support Umbrella Issue


sjperkins:
This issue tracks remaining issues from the TODO list here: https://github.com/google/tensorstore/pull/91#issue-1649705676

Related: 

- https://github.com/google/tensorstore/issues/46
- https://github.com/google/tensorstore/pull/91

# KVStore

- [ ] Versioned Bucket Support. From a high-level perspective would involve combining the `x-amz-version-id` header 
- [ ]  S3 specific rate-limiting (per endpoint?). Current S3 rate limiting is based on the GCS rate limiting code. Relevant documentation
  - https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html
  - https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
  - https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html

# Testing

- [ ] Upgrade to localstack 2.3. `tensorstore::internal::TestKeyValueReadWriteOps` succeeds on `localstack/localstack:latest` image (c.f. https://github.com/localstack/localstack/issues/9076#issuecomment-1711974643) 
- [ ] Install localstack in github actions
- [ ] Benchmarking See https://github.com/google/tensorstore/pull/91#discussion_r1248172227

# Authorisation

- [ ] Credentials from EC2 Metadata Server
- [ ] Credential refresh/timeout functionality (similar to Google Authprovider)

laramiel:
As of https://github.com/google/tensorstore/commit/81aef761d36cff582169fe5b2ab452bcccb6079a we should be in a place where the "//tensorstore/kvstore/s3:localstack_test" will run using a hermetic moto (https://github.com/getmoto/moto) server.

sjperkins:
Thanks, have merged master into #119.

In terms of retry functionality, the following looks like a suitable reference implementation:

- https://github.com/aws/aws-sdk-go/blob/4886ffca0e4f8f35e8abad033e87f0aa305b33cf/aws/client/default_retryer.go#L16
- https://github.com/aws/aws-sdk-go/blob/4886ffca0e4f8f35e8abad033e87f0aa305b33cf/aws/client/default_retryer.go#L78-L123

I might have time to look at this in the second half of November, or, early December.

laramiel:
I have linked "s3" into the default kvstore set; no special changes are required now.

sjperkins:
[...]

PR #119, opened 2023-10-02
Refactor AWS credentials to support credential expiry and greater modularity

sjperkins:
In https://github.com/google/tensorstore/pull/115#pullrequestreview-1640760099, the following code base was pointed out: https://github.com/aws/aws-sdk-go/tree/main/aws/credentials.

I've refactored the AWS credentials to generally follow the referenced patterns and return a `CachedCredentialProvider` which in turn proxies other `CredentialProviders`. See the following Class diagram:

```mermaid
classDiagram
   direction UD
   class AwsCredentialProvider {
     <<Abstract>>
     +GetCredentials() ResultAwsCredentials
     +ExpiresAt() ResultTime
     +IsExpired() bool
   }
   class ExpiryCredentialProvider {

     +Time expiration_
     +Function clock_
     +SetExpiration(Time, Duration) 
   }

   AwsCredentialProvider <|-- ExpiryCredentialProvider
   AwsCredentialProvider <|-- FileCredentialProvider
   AwsCredentialProvider <|-- EnvironmentCredentialProvider
   AwsCredentialProvider <|-- ChainedCredentialProvider
   ExpiryCredentialProvider <|-- EC2CredentialProvider
   AwsCredentialProvider <|-- CachedCredentialProvider
   EnvironmentCredentialProvider  *--  ChainedCredentialProvider
   FileCredentialProvider  *--  ChainedCredentialProvider
   EC2CredentialProvider *-- ChainedCredentialProvider
   ChainedCredentialProvider *-- CachedCredentialProvider
   note for ExpiryCredentialProvider "Implements \nexpiry with a \nTime Variable and\nClock function"
   note for EnvironmentCredentialProvider "Once retrieved,\nnever expires"
   note for FileCredentialProvider "Once retrieved,\nnever expires"
   note for ChainedCredentialProvider  "Iteratively retrieves credentials\nfrom a list of providers\nIsExpiry + ExpiresAt proxies\nprovider which last retrieved creds"
   note for CachedCredentialProvider "Caches Credentials\nRetrieves new credentials on expiry\nGuards encapsulated Provider with a mutex"
```

Then

- CachedCredentialProvider: Credential retrieval protected by a mutex and credentials cached until they expire
  - ChainCredentialProvider: Retrieves the first valid credentials form a list of credentials
    - EnvironmentCredentialProvider (never expires)
    - FileCredentialProvider (never expires)
    - EC2MetadataCredentialProvider (time based expiry, inherits from ExpiryCredentialProvider)

In particular, the ExpiryCredentialProvider implements a clock based expiry mechanism that will primarily be used with the EC2Metadata server.

Other than that, 

- each provider is split into it's own header, source and test case
- Credential provider code has been moved into the `tensorstore/kvstore/s3/credentials` folder.
- Manual IWYU has been applied.

#115 will depend on this, I can merge this code into that PR if you'd prefer to review the changes that way.


laramiel:
Is this intended to replace https://github.com/google/tensorstore/pull/115 or is it more or less part of the changeset where https://github.com/google/tensorstore/pull/115 gets submitted first?

tensorstore/kvstore/s3/credentials/aws_credential_provider.h
```diff
@@ -0,0 +1,75 @@
...
+
+/// Base class for S3 Credential Providers
+///
+/// Implementers should override GetCredentials, IsExpired and,
+/// if the credential source supports the
+/// representation of an expiry date, ExpiresAt.
+class AwsCredentialProvider {
+  public:
+    virtual ~AwsCredentialProvider() = default;
+    virtual Result<AwsCredentials> GetCredentials() = 0;
```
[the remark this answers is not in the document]
sjperkins: The inheritance hierarchy is mostly an attempt to align the C++ implementation with the go implementation. The `AwsCredentialProvider` interface here mostly replicates the go interface here:

https://github.com/aws/aws-sdk-go/blob/b2738d211a2ba65d26731880f30bfdb6abd80858/aws/credentials/credentials.go#L100-L108

and here:

https://github.com/aws/aws-sdk-go/blob/b2738d211a2ba65d26731880f30bfdb6abd80858/aws/credentials/credentials.go#L120-L123

As far as I understand go, the `Expirer` interface is an optional/anonymous interface that exists on the `EC2RoleProvider` [class](https://github.com/aws/aws-sdk-go/blob/b2738d211a2ba65d26731880f30bfdb6abd80858/aws/credentials/ec2rolecreds/ec2_role_provider.go#L40), but not others like the the [env](https://github.com/aws/aws-sdk-go/blob/b2738d211a2ba65d26731880f30bfdb6abd80858/aws/credentials/env_provider.go#L30-L32) provider or the [file](https://github.com/aws/aws-sdk-go/blob/b2738d211a2ba65d26731880f30bfdb6abd80858/aws/credentials/shared_credentials_provider.go#L25-L40) provider.

I can't find the `cached{chain{env, file, ec2role}}` setup in the go code now, might have another look for it in the morning. 

This is why `ExpiresAt` can return `UnimplementedError` to indicate the provider isn't handling expiry (the `CachedCredentialProvider` catches this) and `EC2CredentialProvider` inherits from `ExpiryCredentialProvider`.

Maybe `CachedCredentialProvider` shouldn't really inherit from `AwsCredentialProvider` (the go `Credentials` class/struct doesn't I think, its just a wrapper).

tl;dr. this PR aligns implementations, but maybe its overkill.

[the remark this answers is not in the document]
sjperkins: > Maybe exposing the expiry/refresh is reasonable, but that seems more complicated than returning a Future<>, where the underlying function already checks the expiry.

How would this work? I'm a bit unsure how for example the EC2 metadata expiry date would be retrieved (and set) in the Future. I think the following illustrates my confusion:

```cpp
class EC2Prov {
  public:
    class State { ... };

    absl::Time expiry_;
    absl::Mutex mu_;
    Future<AwsCredentials> creds_; // cached

    Future<AwsCredentials> GetCredentials() {
     absl::WriterReaderLock lock(&mu_);

      if(expiry_ < absl::Now()) { // credentials have expired
        auto [promise, future] = PromiseFuturePair<AwsCredentials>(....);
        auto state = std::IntrusivePtr<State>::Make(std::move(promise));
        future.ExecuteWhenReady()[auto state = std::move(state)]() {
         ...
         auto ec2_json = ...
         expiry_ = ec2_json.ExpiryDate; // This is wrong, but illustrates my confusion
                                                          
         return AwsCredentials{ec2_json.access_key, ec2_json.secret_key, ec2_json.token};
        };
       
       creds_ = future; // Update cached credentials
     }

     return creds_; // Return credentials
   }
};

```

sjperkins:
> Is this intended to replace #115 or is it more or less part of the changeset where #115 gets submitted first? NM, you suggest this happens first. Ok, I'll switch and look at it next.

I think this PR captures the general flow of the go sdk, probably the decision here is whether this is worthwhile, and whether to go with Futures (I'd probably restructure this PR into another PR with futures).

> My comment there was more for reference since I couldn't find the documentation to some of the AWS types; the go code seemed like the easier code to read to discern aws intent...

I've linked the parts of the relevant parts of the go code. Hope this helps

> I'm not sure whether decomposing the responsibilities is very useful when there is really only a single user/implementation of most of these(but I may bias towards functions where others don't).

Are you referring to try (1) `env` (2)  `file` (3)  `ec2` discovery pattern, and then re-use whichever provider returns valid creds? I agree this is the 80% case. Having said that, the ability to chain providers (as in the aws go and c++ sdks) might make auth via other mechanisms easier (SSO?) to implement. I don't know enough about them to comment as to their usefulness.

tensorstore/kvstore/s3/credentials/cached_credential_provider.h
```diff
@@ -0,0 +1,49 @@
...
+#include "tensorstore/kvstore/s3/credentials/aws_credential_provider.h"
+
+namespace tensorstore {
+namespace internal_kvstore_s3 {
+
+/// The Cached Credential Provider encapsulates an instance of AwsCredentialProvider
+/// to provide:
+/// 1. Thread-safe credential retrieval
+/// 2. Caching of credentials
+/// 3. Refresh of credentials upon expiry
```
sjperkins, line 28: So this aligns with `Credential` in the go code

https://github.com/aws/aws-sdk-go/blob/b2738d211a2ba65d26731880f30bfdb6abd80858/aws/credentials/credentials.go#L209-L367

where it proxies a `Provider`. Here, the `AwsCredentialsProvider` is also implemented, mostly so that downstream entities such as the s3 kvstore can take a generic `AwsCredentialsProvider`

tensorstore/kvstore/s3/credentials/aws_credential_provider.h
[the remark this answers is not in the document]
laramiel: In this conceptualization you might keep a shared_ptr<> to some state (could just be the expiry), then you can update that shared_ptr<> in ExecuteWhenReady returns.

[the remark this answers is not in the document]
sjperkins: My concern is more that the expiry check `(expiry < absl::Now())`  would potentially be racy with setting `expiry_` in the `ExecuteWhenReady`. I guess it could be protected with a(nother?) mutex. Even so, in the above example, multiple calls to `GetCredentials` would end up creating multiple futures until `expiry_` was set properly.

The previous example probably isn't very good. Is there an improved way of expressing a Future with an expiry (perhaps somewhere in the codebase).

I also took a look through the `internal::oauth2` code and recalled that there's a `BearerTokenWithExpiration` idea.

https://github.com/google/tensorstore/blob/ce7eb619447818873fcad04f9b85515ad551e276/tensorstore/internal/oauth2/auth_provider.h#L26-L39

Maybe in a similar fashion an `absl::Time expiration` could be added to `AwsCredentials`? This might then 
remove the need for  `AwsCredentials::IsExpired` and `AwsCredentials::ExpiresAt`  methods and could still be used with the Chained and Caching Providers?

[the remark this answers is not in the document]
sjperkins: I didn't figure out how implement a non-racy `Future<AwsCredentials> GetCredentials() { ... }`


> Maybe in a similar fashion an absl::Time expiration could be added to AwsCredentials? This might then
remove the need for AwsCredentials::IsExpired and AwsCredentials::ExpiresAt methods and could still be used with the Chained and Caching Providers?

I experimented with this a bit, but it ended up tending towards what was in master already. If it's desirable I could push up a PR illustrating this.

laramiel:
Is this ready to look at again?

sjperkins:
> Is this ready to look at again?

Yes. master has been merged, including the recent localstack changes.

tensorstore/kvstore/s3/credentials/aws_credential_provider.h
```diff
@@ -48,45 +45,19 @@ struct AwsCredentials {
...
+/// if the credential source supports the
+/// representation of an expiry date, ExpiresAt.
 class AwsCredentialProvider {
- public:
-  virtual ~AwsCredentialProvider() = default;
-  virtual Result<AwsCredentials> GetCredentials() = 0;
+  public:
+    virtual ~AwsCredentialProvider() = default;
+    virtual Result<AwsCredentials> GetCredentials() = 0;
+    virtual bool IsExpired() = 0;
```
laramiel, line 55: Why have an IsExpired() and an ExpiresAt()? stick to a single thing.

tensorstore/kvstore/s3/credentials/file_credential_provider.h
```diff
@@ -0,0 +1,53 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef TENSORSTORE_KVSTORE_S3_FILE_CREDENTIAL_PROVIDER_H
+#define TENSORSTORE_KVSTORE_S3_FILE_CREDENTIAL_PROVIDER_H
+
+#include <string>
+
```
laramiel, line 19: Missing includes.

#include <utility>
#include "absl/time/time.h"

tensorstore/kvstore/s3/credentials/file_credential_provider.h
```diff
@@ -0,0 +1,53 @@
...
+/// from AWS_DEFAULT_PROFILE, AWS_PROFILE before finally defaulting to
+/// "default".
+class FileCredentialProvider : public AwsCredentialProvider {
+ private:
+  // desired profile
+  std::string profile_;
+  bool retrieved_;
+
+ public:
+  FileCredentialProvider(std::string profile)
```
laramiel, line 40: I would expect a FileCredentialProvider could be provided a filename upon construction.

I think that there should also be something like:

Result<std::string> GetAwsCredentialsFileName()

tensorstore/kvstore/s3/credentials/ec2_credential_provider.h
```diff
@@ -19,30 +19,25 @@
...
 #include "tensorstore/util/result.h"
 
 namespace tensorstore {
 namespace internal_kvstore_s3 {
 
 /// Provide S3 credentials from the EC2 Metadata server
-class EC2MetadataCredentialProvider : public AwsCredentialProvider {
+class EC2MetadataCredentialProvider : public ExpiryCredentialProvider {
  public:
   EC2MetadataCredentialProvider(
```
laramiel, line 33: Like FileCredentialProvider, I'd expect an accessor for 

Result<std::string> GetEC2MetadataServiceEndpoint()

And I'd expect to be able to pass an endpoint into the constructor.

laramiel:
Ok, I'm at a point where I can look at this again. I sent a few comments; my current thought is that I like splitting the file, ec2, and environment credentials out into separate, testable files. That's nice and well tested here.  Noted that some includes are missing and the #include guards should also reflect the full file path.

The cached/chained/expiry model I'm not sold on. They provide a mechanism very similar to the registry bits in the old code, so I think that they are useful. But there's a bunch going on there that might be worth changing.

1. If the environment variable exist, then there's no need to chain to other callers.  Once environment credendials are retrieved, they should not change for the life of the process...  There's basically no mechanism in play to refresh them.

2. If file credentials exist, they should also not change for the life of the process.

3. The EC2 credentials should be refersehd every hour or so, sure. Ok.

Should each request of credentials be paired with an expiry? We could return a tuple, and the expiry has moved in and out of the credentials struct.  It's not clear what the better choice is, can you explain what you had in mind there?  I commented elsewhere that I don't find two calls related to expiry useful at this level.

Is it the credential provider job to ensure that the credentials are valid or the callers?  If the provider, then it should accept a time and return a future<>.  If the caller, then there should be a refresh method.

laramiel:
Ok, having looked at this more, I think that I'd like to, for now at least, do the following:

1. Add expires_at to AwsCredentials. 
  - I've thought about the options, and don't like maintaining the expiry separately from the credentials.
  - The GetCredentials call could return a pair<credentials, expiry> or fill in a reference expiry variable,
     but we already have a payload and can just use it.
  - This allows us to minimize the interface to `Result<AwsCredentials> GetCredentials()` (In the future it becomes `Future<AwsCredentials> ...`)

2. Eliminate expiry credentials provider. 
  - With the above it's no longer useful.

3. Merge the generic Chained/Cached credentials provider into a single class, DefaultAwsCredentialsProvider.

4. At the end we're left with this:
  
```
class AwsCredentialProvider {
 public:
  virtual ~AwsCredentialProvider() = default;
  virtual Result<AwsCredentials> GetCredentials() = 0;
};
```  

  
default_aws_credentials_provider.h
```


class DefaultAwsCredentialsProvider : public AwsCredentialProvider {
 public:
   struct Options {
    std::string filename ={};
    std::string profile = {};
    std::shared_ptr<internal_http::HttpTransport> transport;
  };
 
  DefaultAwsCredentialsProvider(Options options) : options_(std::move(options)) {
    credentials_.expires_at = absl::InfinitePast();
  }
  virtual ~DefaultAwsCredentialsProvider() = default;
  Result<AwsCredentials> GetCredentials() override;

 private:
  Options options_;
  absl::Mutex mutex_;
  std::unique_ptr<AwsCredentialProvider> provider_ ABSL_GUARDED_BY(mutex_);
  AwsCredentials credentials_ ABSL_GUARDED_BY(mutex_);
};
```

default_aws_credentials_provider.cc
```
DefaultAwsCredentialProvider::DefaultAwsCredentialProvider() {
  credentials_.expires_at = absl::InfinitePast();
}

Result<AwsCredentials> DefaultAwsCredentialProvider::GetCredentials() {
  {
    absl::ReaderMutexLock lock(&mutex_);
    if (credentials_.expires_at > absl::Now()) {
      return credentials_;
    }
  }

  absl::WriterMutexLock lock(&mutex_);
  // Refresh the existing credentials.
  if (provider_) {
    TENSORSTORE_ASSIGN_OR_RETURN(credentials_, provider_->GetCredentials());
  }

  // Return credentials in this order:
  // 1. AWS Environment Variables, e.g. AWS_ACCESS_KEY_ID
  provider_ = std::make_unique<EnvironmentCredentialProvider>();
  auto credentials_result = provider_->GetCredentials();
  if (credentials_result.ok()) {
    credentials_ = credentials_result.value();
    return credentials_;
  }

  // 2. Shared Credential File, e.g. $HOME/.aws/credentials
  provider_ = std::make_unique<FileCredentialProvider>(options.filename_, options.profile_);
  credentials_result = provider_->GetCredentials();
  if (credentials_result.ok()) {
    credentials_ = credentials_result.value();
    return credentials_;
  }

  // 3. EC2 Metadata Server
  provider_ = std::make_unique<EC2MetadataCredentialProvider>();
  credentials_result = provider_->GetCredentials();
  if (credentials_result.ok()) {
    credentials_ = credentials_result.value();
    return credentials_;
  }
  
  // 4. Anonymous credentials
  provider_ = nullptr;
  credentials_ = AwsCredentials::Anonymous();
  return credentials_;
}

```

sjperkins:
Thanks for the review. I'm on board with the `DefaultCredentialsProvider` and including the expiry in `AwsCredentials`.

> Noted that some includes are missing and the #include guards should also reflect the full file path. I may take those ahead of the other changes.

Regarding the includes, I've been doing this manually which is error-prone. Is there a better way of doing this to avoid back and forth here? I saw https://github.com/google/tensorstore/commit/5ecc6e227652b76ea3b3eba2b6c66db86f416e56 for e.g. and would like to use the same process if automated. I'm aware of:

- https://github.com/include-what-you-use/include-what-you-use which AFAICT would work if I used the cmake build (have been using bazelisk so far)
- https://github.com/storypku/bazel_iwyu

laramiel:
Manually it's error prone, I agree. We have a lot of missing includes scattered around from when our tooling wasn't as good.

We used to use  iwyu, which doesn't actually require CMake to use, just to build. You could try installing a pre-built version, which I think exists for ubuntu: `sudo apt-get install -y iwyu`

Anyway, now we get most of those iwyu changes from a language server which the internal tools team maintains...

sjperkins:
> which AFAICT would work if I used the cmake build (have been using bazelisk so far)

Also, w.r.t. IWYU and cmake: IWYU seems to need a complete list of include dirs passed in for each `.cc` file so I'm thinking it would need include dirs for every third party package bazel downloads and compiles, hence my hypothesis that it'll need the cmake build to properly work. I didn't spend too much time on this route though.

tensorstore/kvstore/s3/credentials/aws_credential_provider.h
```diff
@@ -12,82 +12,57 @@
...
   /// AWS_SESSION_TOKEN
-  std::string session_token;
+  std::string session_token = {};
+  /// Expiration date
+  absl::Time expires_at = absl::InfinitePast();
+
+  /// Anonymous credentials expiring duration seconds from now
+  static AwsCredentials Anonymous(
+      const absl::Duration& duration = absl::InfiniteDuration()) {
+    return AwsCredentials{{}, {}, {}, absl::Now() + duration};
```
laramiel, line 51: absl::InfiniteFuture()

tensorstore/kvstore/s3/credentials/file_credential_provider.cc
```diff
@@ -0,0 +1,147 @@
...
+    result = *credentials_file;
+  } else {
+    if (auto home_dir = GetEnv("HOME"); home_dir) {
+      result = JoinPath(*home_dir, kDefaultAwsCredentialsFilePath);
+    } else {
+      return absl::NotFoundError("Could not read $HOME");
+    }
+  }
+
+  if (auto fstream = std::ifstream(result.c_str()); !fstream.good()) {
```
laramiel, line 68: There's no need to lookup the filename here; an attempt is made to open it below, which will fail appropriately.

tensorstore/kvstore/s3/credentials/file_credential_provider.cc
```diff
@@ -0,0 +1,147 @@
...
+  auto profile = !profile_.empty()
+                     ? std::string(profile_)
+                     : GetEnv(kEnvAwsProfile).value_or(kDefaultProfile);
+
+  auto credentials = AwsCredentials{};
+  auto section_name = std::string{};
+  auto line = std::string{};
+  auto profile_found = false;
+
+  while (std::getline(ifs, line)) {
```
laramiel, line 102: We have used riegeli for a lot of our common file io. In this case it would look like this:

  riegeli::FileReader reader(filename);
  if (!reader.ok()) {
    return reader.status();
  }

  std::string line;
  while (riegeli::ReadLine(reader, line)) {
    // current loop
  }

... though consider this informational, not required at the moment.

tensorstore/kvstore/s3/credentials/file_credential_provider.cc
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  if (!profile_found) {
+    return absl::NotFoundError(absl::StrCat("Profile [", profile,
+                                            "] not found "
+                                            "in credentials file [",
+                                            filename, "]"));
+  }
+
+  ABSL_LOG_FIRST_N(INFO, 1)
```
laramiel, line 139: See my fancy new logging bits if you want verbose logging:

namespace {
    ABSL_CONST_INIT internal_log::VerboseFlag s3_logging("s3");
}

ABSL_LOG_IF(INFO, s3_logging.Level(1)) << "Using ...";

tensorstore/kvstore/s3/credentials/environment_credential_provider.cc
```diff
@@ -0,0 +1,64 @@
...
+static constexpr char kEnvAwsAccessKeyId[] = "AWS_ACCESS_KEY_ID";
+// AWS user password
+static constexpr char kEnvAwsSecretAccessKey[] = "AWS_SECRET_ACCESS_KEY";
+// AWS session token
+static constexpr char kEnvAwsSessionToken[] = "AWS_SESSION_TOKEN";
+
+}  // namespace
+
+Result<AwsCredentials> EnvironmentCredentialProvider::GetCredentials() {
+  if (auto access_key = GetEnv(kEnvAwsAccessKeyId); access_key) {
```
laramiel, line 43: Nit. I prefer unnested / early return in general because I think that it flows better.

auto env = internal::GetEnv...
if (!env) {
  return absl::NotFoundError(...);
}
// other processing

tensorstore/kvstore/s3/credentials/environment_credential_provider.cc
```diff
@@ -0,0 +1,64 @@
...
+
+#include <optional>
+#include <string>
+
+#include "absl/log/absl_log.h"
+#include "absl/status/status.h"
+#include "absl/strings/str_cat.h"
+#include "absl/time/time.h"
+#include "tensorstore/internal/env.h"
+
```
laramiel, line 25: #include "tensorstore/kvstore/s3/credentials/aws_credential_provider.h"
#include "tensorstore/util/result.h"

tensorstore/kvstore/s3/credentials/file_credential_provider.cc
```diff
@@ -0,0 +1,147 @@
...
+
+#include "absl/log/absl_log.h"
+#include "absl/status/status.h"
+#include "absl/strings/ascii.h"
+#include "absl/strings/str_cat.h"
+#include "absl/time/time.h"
+#include "tensorstore/internal/env.h"
+#include "tensorstore/internal/path.h"
+#include "tensorstore/util/result.h"
+
```
laramiel, line 30: #include "tensorstore/kvstore/s3/credentials/aws_credential_provider.h"

tensorstore/kvstore/s3/credentials/BUILD
```diff
@@ -0,0 +1,115 @@
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+load("//bazel:tensorstore.bzl", "tensorstore_cc_library", "tensorstore_cc_test")
+
+package(default_visibility = ["//visibility:public"])
+
+licenses(["notice"])
+
+tensorstore_cc_library(
```
laramiel, line 8: I think that you could break this into separate libraries if you did the following:

1. aws_credential_provider.h has only: AwsCredentials, AwsCredentialProvider.

2. default_credential_provider  adds: 

AwsCredentialProviderFn
RegisterAwsCredentialProviderProvider
GetAwsCredentialProvider

3. Users would mostly need to include awsx_credential_provider; the driver would need to include default_credential_provider as well.

tensorstore/kvstore/s3/validate_test.cc
```diff
@@ -14,6 +14,8 @@
 
 #include "tensorstore/kvstore/s3/validate.h"
 
+#include <memory>
```
laramiel, line 17: this is one of those bad iwyu cases. :/

tensorstore/kvstore/s3/credentials/environment_credential_provider_test.cc
```diff
@@ -0,0 +1,69 @@
...
+
+#include "tensorstore/internal/env.h"
+#include "tensorstore/internal/test_util.h"
+#include "tensorstore/util/future.h"
+#include "tensorstore/util/result.h"
+#include "tensorstore/util/status_testutil.h"
+
+namespace {
+
+using ::tensorstore::Future;
```
laramiel, line 30: Future is unused here, remove the alias and the include.

tensorstore/kvstore/s3/credentials/ec2_credential_provider.h
laramiel: ... this one is still here. For s3-compatible stores we need to be able to set a specific endpoint for the credentials request.

tensorstore/kvstore/s3/credentials/BUILD
```diff
@@ -0,0 +1,115 @@
...
+        "@com_google_absl//absl/status",
+        "@com_google_absl//absl/strings",
+        "@com_google_absl//absl/synchronization",
+        "@com_google_absl//absl/time",
+        ],
+)
+
+
+tensorstore_cc_library(
+  name = "test_utils",
```
laramiel, line 43: Add `testonly = 1,` to restrict use to tests.

tensorstore/kvstore/s3/s3_key_value_store.cc
```diff
@@ -79,14 +80,18 @@
...
-
 // specializations
 #include "tensorstore/internal/cache_key/std_optional.h"  // IWYU pragma: keep
 #include "tensorstore/internal/json_binding/std_array.h"  // IWYU pragma: keep
 #include "tensorstore/internal/json_binding/std_optional.h"  // IWYU pragma: keep
 #include "tensorstore/serialization/fwd.h"  // IWYU pragma: keep
 #include "tensorstore/serialization/std_optional.h"  // IWYU pragma: keep
 #include "tensorstore/util/garbage_collection/std_optional.h"  // IWYU pragma: keep
+#include "tensorstore/internal/json_binding/bindable.h"
+#include "tensorstore/internal/poly/poly_impl.h"
```
laramiel, line 91: iwyu doesn't handle poly well; remove poly/poly_impl / poly/storage  here.

laramiel: In fact, I think that none of these edits are needed.

When running IWYU, double check where include edits happen & keep them in the main-block above, this section is for specific specializations that are ADL-used, not directly used.

tensorstore/kvstore/s3/credentials/default_credential_provider.h
```diff
@@ -0,0 +1,73 @@
...
+/// obtain fresher credentials
+class DefaultAwsCredentialsProvider : public AwsCredentialProvider {
+ public:
+  /// Options to configure the provider. These include the:
+  ///
+  /// 1. Shared Credential Filename
+  /// 2. Shared Credential File Profile
+  /// 3. Http Transport for querying the EC2 Metadata Server
+  struct Options {
+    std::string filename = {};
```
laramiel, line 52: I think that you can leave off the = {} here as std::string is default constructed to an empty string.

tensorstore/kvstore/s3/credentials/default_credential_provider_test.cc
```diff
@@ -0,0 +1,137 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "tensorstore/kvstore/s3/credentials/default_credential_provider.h"
+
+#include <gtest/gtest.h>
+
+#include <fstream>
+#include <string>
+
```
laramiel, line 21: #include <new>
#include "absl/time/time.h"

Also, Future is unused.

SetEnv / UnsetEnv appear to be unused as well. Perhaps a test is missing for those?

laramiel:
Maybe running the clangd language server will give you better suggestions than iwyu... iwyu certainly has some misses. In general you can drop an iwyu-included header if none of the types declared in the header are directly spelled in the source files. Unless the file is adl-used...

For example, highlighting this principle with `Result<T>` and `Future<T>`, the include can be "omitted" if the type isn't explicitly spelled out, such as in the use below:

```
auto x_future = FunctionReturningFuture();
```

If, however, the type is spelled out, then the include is required, as in:

```
Future<int> x_future = FunctionReturningFuture();
```

tensorstore/kvstore/s3/credentials/aws_credential_provider.h
sjperkins: Just for context: The original intent of this PR was to align the implementation with the aws go sdk, which had this sort of interface. This isn't relevant now that we're going with `DefaultCredentialsProvider`,

tensorstore/kvstore/s3/credentials/ec2_credential_provider.h
sjperkins: Thanks for prodding here. The UI tends to collapse and hide older comments which sometimes results in me missing these comments or seeing them later.

tensorstore/kvstore/s3/credentials/default_credential_provider.h
sjperkins: removed

tensorstore/kvstore/s3/s3_key_value_store.cc
sjperkins: removed

tensorstore/kvstore/s3/credentials/BUILD
sjperkins: done

tensorstore/kvstore/s3/credentials/environment_credential_provider_test.cc
sjperkins: removed.

tensorstore/kvstore/s3/validate_test.cc
sjperkins: removed

tensorstore/kvstore/s3/credentials/file_credential_provider.h
sjperkins: Added  a `GetFileName()` accessor

tensorstore/kvstore/s3/credentials/ec2_credential_provider.h
sjperkins: Added a `GetEndpoint()` accessor

tensorstore/kvstore/s3/credentials/aws_credential_provider.h
sjperkins: done

tensorstore/kvstore/s3/credentials/file_credential_provider.cc
sjperkins: removed

tensorstore/kvstore/s3/credentials/environment_credential_provider.cc
sjperkins: I've used the above style in the env cred provider and the file cred provider.

tensorstore/kvstore/s3/credentials/environment_credential_provider.cc
sjperkins: added

tensorstore/kvstore/s3/credentials/file_credential_provider.cc
sjperkins: added

tensorstore/kvstore/s3/credentials/BUILD
sjperkins: This has been separated into the following libraries: 

1. aws_credential_provider
2. default_credentials_provider
3. environment_credentials_provider
4. file_credentials_provider
5. ec2_credentials_provider

tensorstore/kvstore/s3/credentials/file_credential_provider.h
sjperkins: I think this doesn't apply anymore due to the refactor of each provider into separate libraries.

tensorstore/kvstore/s3/credentials/default_credential_provider_test.cc
sjperkins: added `<new>` and `time.h`, removed Future. Added tests that used SetEnv and UnsetEnv.

As an aside, clangd thinks `<new>` should be removed!

tensorstore/kvstore/s3/credentials/file_credential_provider.cc
sjperkins: Most unfortunately I didn't have time to look at this in this PR. I've added an item https://github.com/google/tensorstore/issues/114 to review this.

sjperkins:
> Maybe running the clangd language server will give you better suggestions than iwyu... iwyu certainly has some misses. In general you can drop an iwyu-included header if none of the types declared in the header are directly spelled in the source files. Unless the file is adl-used...

I installed:

- clangd 17
- https://github.com/hedronvision/bazel-compile-commands-extractor to create the `compile_commands.json`

and configured clangd as follows: 

```
If:
  PathMatch: tensorstore/kvstore/s3/*
Diagnostics:
  UnusedIncludes: Strict
  MissingIncludes: Strict
```

but, while it did seem to be fairly good at spotting unused includes, it didn't seem to detect missing includes all that well. I've applied best effort here but its still been a very manual process. I may need to spend more time working on configuring it, maybe on the weekend.

laramiel:
Can you update the change description to better reflect the current state? I'll import and see what else I need; maybe only the updated description.

sjperkins:
> Can you update the change description to better reflect the current state? 

I've done so.

laramiel:
Ok, it looks like that didn't squash properly. Well next time. :)

sjperkins:
> Ok, it looks like that didn't squash properly

Maybe because there wasn't a master merge for this https://github.com/google/tensorstore/commit/daed8c2bb61d6315243b46e32df3a22b20f87eb3. 

Squashing the PR into one commit may be a viable future strategy.

PR #127, opened 2023-11-13
Structarray

blasscoc-tgs:


tensorstore/driver/kvs_backed_chunk_driver.h
```diff
@@ -740,6 +743,9 @@ class OpenState : public MetadataOpenState {
   /// If the `metadata` is not compatible, returns an error.
   virtual Result<size_t> GetComponentIndex(const void* metadata,
                                            OpenMode open_mode) = 0;
+  /// attempt to cast to a Void type ...
+  virtual Result<std::shared_ptr<void>> AsByteArray(
```
jbms, line 747: I think we can make this change purely in the zarr driver, rather than here in kvs_backed_chunk_driver.h

In particular, if the user does not specify a field, then we can always just open as a byte array.

I also think rather than create synthetic metadata it would be better to just create an additional special dtype field for the byte array representation.  The synthetic metadata approach has some drawbacks.

Issue #118, opened 2023-09-27
Python-implementable kvstores


normanrz:
In a recent meeting, @jbms, I and others discussed that it would be useful to expose a Python interface to implement custom kvstores. This would allow library users to implement their own storage backends such as S3, Azure, Webdav more easily. It could also enable an fsspec adapter.

PR #134, opened 2024-01-29
Enable grpc_kvstore for building

zlobober:
Hi!

I am working on integrating tensorstore with [YTsaurus](https://github.com/ytsaurus/ytsaurus) as an underlying storage. 
In order to obtain the PoC I used the grpc_kvstore implementation of kvstore. It looks like this implementation is not currently considered "stable" or "officially supported", which is perfectly ok for me as our server-side implementation is also fresh and rapidly changing.

Still, not being able to use it with official pypi package of tensorstore is not very convenient as we have to provide patched version of tensorstore package with the only difference from this PR. What do you think about enabling use of this kvstore (possibly, with a number of disclaimers about being experimental and unstable)?

If you agree on this, we would be glad to announce (at some point of time) that YTsaurus provides integration with tensorstore via a certain grpc service. It may then happen that I'll be bringing more PRs extending the functionality of grpc_kvstore, as now it seems that it does not cover all the functionality from the internal kvstore [interface](https://github.com/google/tensorstore/blob/master/tensorstore/kvstore/operations.h).

Cheers,
Maxim.

### DIFF
--- a/tensorstore/kvstore/BUILD
+++ b/tensorstore/kvstore/BUILD
@@ -33,7 +33,7 @@ DRIVERS = DRIVER_DOCS + [
     "gcs_http",
 ]
 
-EXTRA_DRIVER_TARGETS = []
+EXTRA_DRIVER_TARGETS = ["//tensorstore/kvstore/grpc:grpc_kvstore"]
 
 tensorstore_cc_library(
     name = "all_drivers",


jbms:
The current grpc kvstore driver serves as an example, but is not really intended for production use as it lacks support for authentication, and we also do not guarantee stability of the protobuf interface.

Potentially these issues could be addressed, though.

Another alternative would be for tensorstore to communicate with YTsaurus using a different, already established protocol, like S3 or the GCS grpc protocol.  Does YTsaurus already support other already-established network protocols?

zlobober:
Thanks for your response. I already spent some time on deciding how to act, so let my share my thoughts. Sorry for a long response.

In general, I see somewhat like 4 ways of integrating tensorstore with YTsaurus.

First way is **vendoring**. It requires writing custom kvstore implementation specifically for YTsaurus which implements necessary C++ interfaces, and if I were you I would not want some third-party client for not yet really popular backend in your repository. 

Second way is **run-time plugin**; it would require fixing some ABI and letting third-party kv-stores be plugged into tensorstore in runtime, allowing, for example, potential kvstore plugins to be distributed as separate Python packages. It would require fixing C/C++ ABI, which sounds like a much harder and painful procedure, comparing to fixing a grpc interface.

Third one is, just as you said, **mimicking** for an existing tensorstore backend. Let me consider S3 API as an example, but I believe that the same reasoning must be applicable for GCS grpc API. Despite being in general a good direction from a product standpoint for us, this is a really long way to follow; S3 API has lots and lots of methods with complex behavior.

Let's also note that tensorstore uses a small portion of S3 API, and that some parts of S3 API might be even limiting for support in YTsaurus; for example, YTsaurus provides transactional model with strong isolation properties, while S3 does not. Tensorstore might benefit from using YTsaurus own API, rather than restricting to a common part, that is expressible both in YTsaurus and S3.

At the same time, implementing the bare minimum needed for tensorstore looks like a not very good compromise, as it creates a false sense of S3-compatibility (or GCS-compatibility) for our potential users. Moreover, it makes even vaguer, what is the actual contract between tensorstore and YTsaurus. When somebody decides to use some new method of S3 API, he would be somewhat "right" despite the fact YTsaurus may not support that method at that moment of time.

Now, let's evaluate the final option of using **grpc** kvstore. It provides users of tensorstore with the way how to write backends on various languages, not only C++ (cf. #118). It also mitigates the issue that SDK for a particular actual underlying system may be really hard to vendor, which is the case for our own C++ SDK -- it requires a certain rare build system.

My personal view on the current state of [operations.h](https://github.com/google/tensorstore/blob/master/tensorstore/kvstore/operations.h) is that the interface of kv-store looks mature enough to assume that it will not be radically refactored in the future. This interface is expressible in grpc in a quite straightforward manner, which is proven by the current version of the grpc protocol (aside from it being a little bit stale since it was not updated since Mar 2023).

Regarding the lack of authentication -- you are right. Funny enough, this is not an issue in my particular case, because in my PoC the grpc server, which is called "tensorproxy", acts as a lightweight proxy, bound to some port at localhost. It is intended to run close to a process hosting the tensorstore library.  The "tensorproxy" process takes responsibility for performing authenticated calls to YTsaurus, e.g. it requires YTsaurus user credentials as a configuration for running. Still, supporting some sort of authentication here is a good idea, as it would enable me to move "tensorproxy" from being essentially the part of the client-side to being a part of the server-side.

I agree with the fact that grpc interface requires some improvements. I'd like to suggest some compromise here: if you, at the very least, expose the current grpc_kvstore driver (possibly renaming it into "expreimental_grpc_kvstore"), we would gladly use it despite the interface being unstable, and rewrite our grpc service accordingly whenever needed. After all, it took only a couple of hours of my time to implement the current grpc interface from scratch, so I'll be able to adapt the interface at a low price in the future. 

It may even happen that we contribute some additional features missing in the current grpc_kvstore interface (in comparison to the internal interface in operations.h); but in order to experiment with the actual performance or feature-set, we would like to start using it without having to privately patch your library.

jbms:
Thanks for your detailed explanation.  I agree with your analysis that the grpc kvstore provides the best solution for your use case --- we will have to decide whether to rename it to experimental but will try to get it enabled by default shortly.

zlobober:
Hi @jbms,
any updates on this topic? We are ready to bring first pull requests further improving the feature set of grpc_kvstore, should we bring them on top of current unexposed version of code, or it is better to wait until grpc_kvstore becomes enabled for everybody?

laramiel:
@jbms and I discussed this, and we're planning on enabling the driver relatively soon. However before I enable the grpc driver globally I feel like it's important to add retry logic for failed rpcs, as anecdotally we see a lot of test failures related to this, which need to be fixed before we enable by default.

Also, we need to consider whether changing GRPC_ARG_MAX_RECEIVE_MESSAGE_LENGTH is sufficient for our uses, or whether we need to convert to streaming rpcs for read/write.  I suspect that we will need to use streaming gRPC calls.

PR #149, opened 2024-04-02
Use AWS C++ SDK for communicating with S3

sjperkins:
Based on this discussion:

- https://github.com/google/tensorstore/issues/114#issuecomment-1936418508

TODO:

- ~~[x] Integrate AWS CPP SDK #defines into a `config.h` file variant. https://github.com/google/tensorstore/pull/149#discussion_r1548284979~~


third_party/com_github_aws_cpp_sdk/aws_cpp_sdk.BUILD.bazel
```diff
@@ -0,0 +1,180 @@
...
+
+exports_files(["LICENSE"])
+
+cc_library(
+    name = "core",
+    srcs = glob([
+        "aws-cpp-sdk-core/source/*.cpp",  # AWS_SOURCE
+        "aws-cpp-sdk-core/source/external/tinyxml2/*.cpp",  # AWS_TINYXML2_SOURCE
+        "aws-cpp-sdk-core/source/external/cjson/*.cpp",  # CJSON_SOURCE
+        "aws-cpp-sdk-core/source/auth/*.cpp",  # AWS_AUTH_SOURCE
```
sjperkins, line 16: Naively, I assume the challenge will be to see how much of the source can be excised, while retaining the auth capalities.

sjperkins (commented):
I've included the tensorflow io deps into bazel. The bulk of the work needs to be done, but  it's probably worth reviewing how the dependency representation has been done as I'm not very experienced with bazel. I've tried to replicate the tensorflow io deps with tensorstore's workplace flavour.

FWIW it builds locally and hopefully the dummy s3 dep I've added to the s3_request_builder works on CI.

third_party/com_github_aws_c_common/aws_c_common.BUILD.bazel
```diff
@@ -0,0 +1,46 @@
...
+exports_files(["LICENSE"])
+
+cc_library(
+    name = "aws-c-common",
+    srcs = glob([
+        "include/aws/common/*.h",
+        "include/aws/common/private/*.h",
+        "source/*.c",
+    ]) + select({
+        "@bazel_tools//src/conditions:windows": glob([
```
laramiel, line 17: "@platforms//os:windows":

third_party/com_github_aws_c_common/system.BUILD.bazel
```diff
@@ -0,0 +1,4 @@
+cc_library(
```
laramiel, line 1: We probably don't need the system.BUILD.bazel files here; since it's c++, and since there are not common packages, we will likely build from source for now.  Though maybe at some point conda-forge will want this.

https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/sdk-from-pm.html

third_party/com_github_aws_c_common/workspace.bzl
```diff
@@ -0,0 +1,34 @@
...
+def repo():
+    maybe(
+        third_party_http_archive,
+        name = "com_github_aws_c_common",
+        sha256 = "01c2a58553a37b3aa5914d9e0bf7bf14507ff4937bc5872a678892ca20fcae1f",
+        strip_prefix = "aws-c-common-0.4.29",
+        urls = [
+            "https://storage.googleapis.com/mirror.tensorflow.org/github.com/awslabs/aws-c-common/archive/v0.4.29.tar.gz",
+            "https://github.com/awslabs/aws-c-common/archive/v0.4.29.tar.gz",
+        ],        build_file = Label("//third_party:com_github_aws_c_common/aws_c_common.BUILD.bazel"),
```
laramiel, line 27: nit. formatting.

third_party/com_github_aws_checksums/aws_checksums.BUILD.bazel
```diff
@@ -0,0 +1,42 @@
...
+cc_library(
+    name = "aws-checksums",
+    srcs = glob([
+        "include/aws/checksums/*.h",
+        "include/aws/checksums/private/*.h",
+        "source/*.c",
+    ]) + [
+        "crc_hw.c",
+    ],
+    hdrs = [],
```
laramiel, line 19: I would expect a header file. Is that not the case?

third_party/com_github_aws_cpp_sdk/aws_cpp_sdk.BUILD.bazel
```diff
@@ -0,0 +1,180 @@
...
+genrule(
+    name = "SDKConfig_h",
+    outs = [
+        "aws-cpp-sdk-core/include/aws/core/SDKConfig.h",
+    ],
+    cmd = "\n".join([
+        "cat <<'EOF' >$@",
+        "#define USE_AWS_MEMORY_MANAGEMENT",
+        "#if defined(_MSC_VER)",
+        "#include <Windows.h>",
```
laramiel, line 175: We might want to add something like:

#ifndef WIN32_LEAN_AND_MEAN
#define WIN32_LEAN_AND_MEAN
#endif

before #include <Windows.h>

third_party/com_github_aws_cpp_sdk/aws_cpp_sdk.BUILD.bazel
laramiel: We now have tinyxml2, so  it would be good to remove that if we can. 

:)

third_party/com_github_aws_cpp_sdk/aws_cpp_sdk.BUILD.bazel
```diff
@@ -0,0 +1,180 @@
...
+        "aws-cpp-sdk-core/include/aws/core/utils/memory/stl/*.h",  # UTILS_STL_HEADERS
+        "aws-cpp-sdk-core/include/aws/core/utils/logging/*.h",  # UTILS_LOGGING_HEADERS
+        "aws-cpp-sdk-core/include/aws/core/utils/ratelimiter/*.h",  # UTILS_RATE_LIMITER_HEADERS
+        "aws-cpp-sdk-core/include/aws/core/utils/stream/*.h",  # UTILS_STREAM_HEADERS
+        "aws-cpp-sdk-core/include/aws/core/external/cjson/*.h",  # CJSON_HEADERS
+        "aws-cpp-sdk-core/include/aws/core/external/tinyxml2/*.h",  # TINYXML2_HEADERS
+        "aws-cpp-sdk-core/include/aws/core/http/curl/*.h",  # HTTP_CURL_CLIENT_HEADERS
+        "aws-cpp-sdk-core/include/aws/core/utils/crypto/openssl/*.h",  # UTILS_CRYPTO_OPENSSL_HEADERS
+    ]),
+    defines = [
```
laramiel, line 78: Do these not go into a config.h file or similar?

third_party/com_github_aws_cpp_sdk/workspace.bzl
```diff
@@ -0,0 +1,38 @@
...
+# limitations under the License.
+
+load("//third_party:repo.bzl", "third_party_http_archive")
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
+
+def repo():
+    maybe(
+        third_party_http_archive,
+        name = "com_github_aws_cpp_sdk",
+        patch_cmds = [
```
laramiel, line 22: Nice patch.  Have you tried this with the cmake build?

third_party/com_github_aws_checksums/aws_checksums.BUILD.bazel
sjperkins: No headers were specified in the original: https://github.com/tensorflow/io/blob/4d70341df05ddce09c257d4ecaaaba1873594105/third_party/aws-checksums.BUILD#L10-L25

third_party/com_github_aws_cpp_sdk/aws_cpp_sdk.BUILD.bazel
sjperkins: I'm not sure. This was copied straight out of https://github.com/tensorflow/io/blob/4d70341df05ddce09c257d4ecaaaba1873594105/third_party/aws-sdk-cpp.BUILD#L78-L94

I'd rather spend time on this after the feasibility of trimming down the AWS SDK to the Auth code has been demonstrated. Added a TODO in the first comment.

third_party/com_github_aws_cpp_sdk/workspace.bzl
sjperkins: I copied it from here https://github.com/tensorflow/io/blob/4d70341df05ddce09c257d4ecaaaba1873594105/WORKSPACE#L224-L236

Yeah this falls over on `cmake -S . -B build` with the following truncated output:

```bash
- verifying file...
       file='/home/simon/code/tensorstore/build/_deps/aws_cpp_sdk-subbuild/aws_cpp_sdk-populate-prefix/src/1.8.187.tar.gz'
-- Downloading... done
-- extracting...
     src='/home/simon/code/tensorstore/build/_deps/aws_cpp_sdk-subbuild/aws_cpp_sdk-populate-prefix/src/1.8.187.tar.gz'
     dst='/home/simon/code/tensorstore/build/_deps/aws_cpp_sdk-src'
-- extracting... [tar xfz]
-- extracting... [analysis]
-- extracting... [rename]
-- extracting... [clean up]
-- extracting... done
[ 33%] No update step for 'aws_cpp_sdk-populate'
[ 44%] Performing patch step for 'aws_cpp_sdk-populate'
sed: -e expression #1, char 1: unknown command: `''
gmake[2]: *** [CMakeFiles/aws_cpp_sdk-populate.dir/build.make:118: aws_cpp_sdk-populate-prefix/src/aws_cpp_sdk-populate-stamp/aws_cpp_sdk-populate-patch] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/aws_cpp_sdk-populate.dir/all] Error 2
gmake: *** [Makefile:91: all] Error 2

CMake Error at /usr/share/cmake-3.29/Modules/FetchContent.cmake:1714 (message):
  Build step for aws_cpp_sdk failed: 2
Call Stack (most recent call first):
  /usr/share/cmake-3.29/Modules/FetchContent.cmake:1854:EVAL:2 (__FetchContent_directPopulate)
  /usr/share/cmake-3.29/Modules/FetchContent.cmake:1854 (cmake_language)
  /usr/share/cmake-3.29/Modules/FetchContent.cmake:2081 (FetchContent_Populate)
  build/third_party/CMakeLists.txt:3 (FetchContent_MakeAvailable)
```

The contents of the patch file (aws_cpp_sdk-populate-patch-info.txt) look like extraneous `;` characters have been added in place of spaces in the `patch_cmds`.

```bash
~/code/tensorstore/build/_deps/aws_cpp_sdk-subbuild/aws_cpp_sdk-populate-prefix/src/aws_cpp_sdk-populate-stamp$ cat aws_cpp_sdk-populate-patch-info.txt 
# This is a generated file and its contents are an internal implementation detail.
# The update step will be re-executed if anything in this file changes.
# No other meaning or use of this file is supported.

command=sed;-i.bak;'s/UUID::RandomUUID/Aws::Utils::UUID::RandomUUID/g';aws-cpp-sdk-core/source/client/AWSClient.cpp;&&;sed;-i.bak;'s/__attribute__;(;(;visibility;(;default;);););//g';aws-cpp-sdk-core/include/aws/core/external/tinyxml2/tinyxml2.h;&&;/usr/bin/cmake;-E;copy;/home/simon/code/tensorstore/third_party/com_github_aws_cpp_sdk/aws_cpp_sdk.BUILD.bazel;BUILD.bazel;&&;/usr/bin/cmake;-E;copy;/home/simon/code/tensorstore/build/third_party/aws_cpp_sdk-proxy-CMakeLists.txt;CMakeLists.txt
work_dir=/home/simon/code/tensorstore/build/_deps/aws_cpp_sdk-src
```

Any insights here? Maybe `http_archive` does some string handling that `third_party_http_archive` does not yet?

third_party/com_github_aws_c_common/aws_c_common.BUILD.bazel
sjperkins: Changed

third_party/com_github_aws_checksums/aws_checksums.BUILD.bazel
```diff
@@ -0,0 +1,42 @@
...
+#   AWS CheckSums
+
+package(default_visibility = ["//visibility:public"])
+
+licenses(["notice"])  # Apache 2.0
+
+exports_files(["LICENSE"])
+
+cc_library(
+    name = "aws-checksums",
```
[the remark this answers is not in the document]
sjperkins: Fixed.

third_party/com_github_aws_cpp_sdk/aws_cpp_sdk.BUILD.bazel
sjperkins: Indeed, lets see whats possible here.

sjperkins: It looks like the `tinyxml2` vendored into the AWS c++ sdk has been patched. Quoting from  https://github.com/aws/aws-sdk-cpp/blob/d79ece080eba11212b10a857ae551a4b8423fe00/src/aws-cpp-sdk-core/include/aws/core/external/tinyxml2/tinyxml2.h#L23-L28.

```
  (1) Memory management operations use aws memory management api
  (2) Import-export preprocessor logic tweaked for better integration into core library
  (3) Wrapped everything in Amazon namespace to prevent static linking issues if the user includes a version of this code through another dependency
```

Would it be reasonable to use the AWS C++ SDK version of tinyxml2, instead of the vanilla version currently used with tensorflow?

sjperkins: > Naively, I assume the challenge will be to see how much of the source can be excised, while retaining the auth capalities.

Naively. So everything in the core library is tightly coupled. It doesn't seem possible to easily excise portions of the core library out. However, here are the `.so` binary sizes after stripping:

```bash
$ ~/code/tensorstore/bazel-bin/external$ ls -lh com_github_aws_cpp_sdk/libcore.so
-rwxr-xr-x 1 simon simon 3.8M Apr  3 15:42 com_github_aws_cpp_sdk/libcore.so

$ ls -lh com_github_aws_c_common/libaws_c_common.so
-rwxr-xr-x 1 simon simon 193K Apr  3 15:40 com_github_aws_c_common/libaws_c_common.so

$ ls -lh com_github_aws_c_event_stream/libaws_c_event_stream.so
-rwxr-xr-x 1 simon simon 35K Apr  3 15:40 com_github_aws_c_event_stream/libaws_c_event_stream.so

$ ls -lh com_github_aws_checksums/libaws_checksums.so
-rwxr-xr-x 1 simon simon 43K Apr  3 15:40 com_github_aws_checksums/libaws_checksums.so
```

Lets round that up and say that it costs `~4.1MB total` to get the AWS SDK code functionality. Considering that some of the functionality is already replicated already in S3RequestBuilder, depending on `core` may not be a bad tradeoff in terms of code maintenance.

jbms: In most cases outside tests the AWS libraries would be statically linked to tensorstore and the linker may automatically remove most unused code anyway.

sjperkins: > In most cases outside tests the AWS libraries would be statically linked to tensorstore and the linker may automatically remove most unused code anyway.

Yes, the `.so` sizes would only establish a rough upper bound on include size. I'm trying to estimate before making significant code changes, but I guess including the code and inspecting the final wheel (for e.g.) is the only way to be sure.

`libs3.so` is `9.2MB` stripped.

sjperkins:
So a first order strategy here might be:


- Replace the current AWS S3 Auth code with a [Aws::Auth::DefaultCredentialProviderChain](https://sdk.amazonaws.com/cpp/api/LATEST/aws-cpp-sdk-core/html/class_aws_1_1_auth_1_1_default_a_w_s_credentials_provider_chain.html).

and in `S3RequestBuilder::BuildRequest`:

- Create a dummy [Aws::Http::StandardHttpRequest](https://sdk.amazonaws.com/cpp/api/LATEST/aws-cpp-sdk-core/html/class_aws_1_1_http_1_1_standard_1_1_standard_http_request.html). 
  - `Aws::String` objects are simply typedef `std::basic_strings` so that shouldn't be an issue.
  - An Adapter will be needed to represent the `absl::Cord` payload as an `Aws::Iostream`.  
- Sign the `Aws::Http::StandardHttpRequest` with an [AwsAuthV4Signer](https://sdk.amazonaws.com/cpp/api/LATEST/aws-cpp-sdk-core/html/class_aws_1_1_client_1_1_a_w_s_auth_v4_signer.html#pub-methods) using the `Aws::Auth::DefaultCredentialProviderChain`.
- Get the `Authorization` Header off the `Aws::Http::StandardHttpRequest` and use in `tensorstore::http::HttpRequest`

I guess at some point the [S3 Encryption](https://sdk.amazonaws.com/cpp/api/LATEST/aws-cpp-sdk-s3-encryption/html/class_aws_1_1_s3_encryption_1_1_s3_encryption_client.html) API will become relevant.

third_party/com_github_aws_cpp_sdk/aws_cpp_sdk.BUILD.bazel
sjperkins: I've moved this defines into a `write_file` rule to generate `aws-cpp-sdk-core/include/aws/core/SDKConfig.h`, with some platform specific defines.

third_party/com_github_aws_cpp_sdk/workspace.bzl
sjperkins: `patch_cmds` was converted into a `.diff file` and applied with `patch`.

sjperkins:
Still WIP, none of the existing codepaths have been touched.

sjperkins:
Hello, just checking in here. I've been reading more of the AWS C++ SDK and Tensorstore code base.

It's probably possible to provide Adapters for the AWS C++ SDK classes in order to adapt them for use by the tensorstore-equivalent implementation:

- `Aws::Http::HttpRequest`
- `Aws::Http::HttpResponse`
- `Aws::Http::HttpClient`
- `Aws::Utils::Threading::Executor`

This would probably allow the use of the various models `S3Client`, for example

- https://sdk.amazonaws.com/cpp/api/LATEST/aws-cpp-sdk-s3/html/class_aws_1_1_s3_1_1_model_1_1_head_bucket_request.html
- https://sdk.amazonaws.com/cpp/api/LATEST/aws-cpp-sdk-s3/html/class_aws_1_1_s3_1_1_model_1_1_put_object_request.html

See below for the idea with  

```cpp
  class CustomExecutor : public Aws::Utils::Threading::Executor {
  protected:
    // Compiles, but only demonstrates the idea
    // Defer to tensorstore executor
    bool SubmitToThread(std::function<void()> && fn) override {
      auto task = ::tensorstore::ExecutorTask(fn);
      auto bound_fn = ::tensorstore::WithExecutor(..., task);
      ...
      return true;
    }
  };

  // In s3_context.cc
  // Override Aws::Http::HttpClient during AWS setup, with the intention of using
  // tensorflow's HttpClient (and curl implementation)
  // Aws::Http::SetHttpClientFactory(Aws::MakeShared<CustomHttpFactory>(kAWSTag));
  auto ctx = GetAwsContext();

  auto cfg = Aws::Client::ClientConfiguration();
  // Override the default AWS executor with tensorflow's Executor
  cfg.executor = Aws::MakeShared<CustomExecutor>(kAWSTag);

  auto client = Aws::S3::S3Client(cfg);
  // Inherits from Aws::Http::HttpRequest
  auto head_bucket = Aws::S3::Model::HeadBucketRequest().WithBucket("bucket");
  // Under the hood, submits Aws::Http::HttpRequest to CustomHttpClient
  // which converts to a tensorstore::internal_http::HttpRequest and submits it on
  // a tensorstore::Executor
  auto outcome = client.HeadBucket(head_bucket);
  if(!outcome.IsSuccess()) {
    auto & err = outcome.GetError();
    std::cerr << "Error: " << err.GetExceptionName() << ": " << err.GetMessage() << std::endl;
  } else {
    std::cout << "Success" << std::endl;
  }
 ```

The benefit of this approach is that most of the S3 Request logic such as 

- request construction
- endpoint computation
- xml response handling (maybe, haven't dug into the responses yet)

will be handled by the AWS C++ SDK, while the I/O would be performed by tensorstore's implementations.  But this approach might be deemed too heavyweight in terms of code inclusion. I would like to know if there are concerns here before pursuing it further.

**Comment by sjperkins** [comment]
Some incremental progress on the S3Client approach:

- Overrides default SDK
  -  HttpFactory to create a custom HttpClient
  - `CustomHttpClient` converts AWS HttpRequest into tensorstore HttpRequests for submission on tensorstore's curl implementation
  - Conversely, converts tensorstore HttpResponses to AWS HttpResponses
  - S3Client can be configured to use an Executor wrapping a tensorstore executor
- Some basic test cases interacting with localstack that demonstrate use of S3Client with Sync and Async Request styles

Further investigation required:
 - SDK uses `Aws::IOStream` used for Http payloads and there's  expensive conversion between Cord and Aws::`IOStream`. `Aws::IOStream` should probably be modified to encapsulate an `abseil::Cord`.
 - `CustomHttpClient` waits for a tensorstore Future to complete. Ideally, I would like to find some way to yield to the calling tensorstore Executor, but I'm not sure if this is possible. I'll mark the relevant location.

File: tensorstore/kvstore/s3_sdk/s3_context.cc [hunk 10]
```diff
@@ -0,0 +1,294 @@
...
+    const std::shared_ptr<AwsHttpRequest> & request,
+    AwsRateLimiterInterface* readLimiter = nullptr,
+    AwsRateLimiterInterface* writeLimiter = nullptr) const override {
+      // Issue the wrapped HttpRequest on a tensorstore executor
+      auto transport = ::tensorstore::internal_http::GetDefaultHttpTransport();
+      auto [ts_request, payload] = FromAwsRequest(request);
+      ABSL_LOG(INFO) << ts_request << " " << payload;
+      auto future = transport->IssueRequest(ts_request, IssueRequestOptions(payload));
+      // TODO: if possible use a continuation (future.ExecuteWhenReady) here
+      auto response = future.value();
```
sjperkins, line 149: Generally speaking, it would seem desirable to yield to the calling tensorstore Executor here, but I'm not sure that this is really possible.

laramiel:
I started looking at this again yesterday.  Based on the HEAD repositories in github it appears that we need the following libraries imported:

```
"https://github.com/awslabs/aws-c-http",
"https://github.com/awslabs/aws-c-io",
"https://github.com/awslabs/aws-c-auth",
"https://github.com/awslabs/aws-c-mqtt",
"https://github.com/awslabs/aws-c-cal",
"https://github.com/awslabs/aws-c-compression",
"https://github.com/awslabs/aws-checksums",
"https://github.com/awslabs/aws-c-sdkutils",
"https://github.com/awslabs/aws-c-common",
"https://github.com/awslabs/aws-c-event-stream",
"https://github.com/awslabs/aws-c-s3",
"https://github.com/awslabs/aws-crt-cpp",
```

Maybe these:
```
"https://github.com/aws/s2n-tls",
"https://github.com/aws/aws-lc",
```

And then the C++ sdk has a bunch of sub-libraries.
```
"https://github.com/aws/aws-sdk-cpp",

"aws-cpp-sdk-core",
"aws-cpp-sdk-s3",
```

sjperkins:
Thanks, I'll take a look.

sjperkins:
```
"https://github.com/awslabs/aws-c-http",
"https://github.com/awslabs/aws-c-io",
"https://github.com/awslabs/aws-c-auth",
"https://github.com/awslabs/aws-c-mqtt",
"https://github.com/awslabs/aws-c-cal",
"https://github.com/awslabs/aws-c-compression",
"https://github.com/awslabs/aws-checksums",
"https://github.com/awslabs/aws-c-sdkutils",
"https://github.com/awslabs/aws-c-common",
"https://github.com/awslabs/aws-c-event-stream",
"https://github.com/awslabs/aws-c-s3",
"https://github.com/awslabs/aws-crt-cpp",
"https://github.com/aws/s2n-tls",
```

```
"https://github.com/aws/aws-sdk-cpp",

"aws-cpp-sdk-core",
"aws-cpp-sdk-s3",
```

Given that the above are all built with cmake, it may be worth using https://github.com/bazelbuild/rules_foreign_cc to integrate them into bazel, even though `rule_foreign_cc` isn't officially supported.  Let me know if this isn't feasible for some reason.

laramiel:
I don't think that we want rules_foreign_cc; internally we will need to build with bazel anyway.

sjperkins:
> I don't think that we want rules_foreign_cc; internally we will need to build with bazel anyway.

I see.  Based on a brief look through `aws-c-common` HEAD (only), one concern was generating the `config.h.in` file

- https://github.com/awslabs/aws-c-common/blob/main/include/aws/common/config.h.in

In a strict sense, it seems that a pure bazel approach would need to customise the above  for combinations of different platforms and architectures, for each dependency, perhaps through the use of  `write_file` as done here: https://github.com/google/tensorstore/pull/149/files#diff-cee7f961ceb075b160fbdd37a847f8f68be0872f932924267589a78d4dbe8c57R175-R203

This seems doable for one repo. One concern is that a trial and error approach to this might introduce subtle bugs into the SDK build (e.g. builds with newer instructions sets, although these could be disabled for the sake of safety).

However, I may be overthinking this. Did the configs of the above dependencies require extensive customisation in the bazel build mentioned here?

- https://github.com/aws/aws-sdk-cpp/issues/3002#issuecomment-2181754016

laramiel:
Yes. There are only, I think, about 2 config.h.in style files in the aws repositories that we need, this one in aws_c_common and another in aws_cpp_sdk/aws_cpp_sdk_core or something.

I think that the starting point for c-common is something like

```
# Configured for Linux x86_64.
_SUBS = {
    "#cmakedefine AWS_HAVE_GCC_OVERFLOW_MATH_EXTENSIONS": "#define AWS_HAVE_GCC_OVERFLOW_MATH_EXTENSIONS",
    "#cmakedefine AWS_HAVE_GCC_INLINE_ASM": "#define AWS_HAVE_GCC_INLINE_ASM",
    "#cmakedefine AWS_HAVE_MSVC_INTRINSICS_X64": "",
    "#cmakedefine AWS_HAVE_POSIX_LARGE_FILE_SUPPORT": "#define AWS_HAVE_POSIX_LARGE_FILE_SUPPORT",
    "#cmakedefine AWS_HAVE_EXECINFO": "#define AWS_HAVE_EXECINFO",
    "#cmakedefine AWS_HAVE_WINAPI_DESKTOP": "",
    "#cmakedefine AWS_HAVE_LINUX_IF_LINK_H": "#define AWS_HAVE_LINUX_IF_LINK_H",
    "#cmakedefine AWS_HAVE_AVX2_INTRINSICS": "#define AWS_HAVE_AVX2_INTRINSICS",
    "#cmakedefine AWS_HAVE_AVX512_INTRINSICS": "#define AWS_HAVE_AVX512_INTRINSICS",
    "#cmakedefine AWS_HAVE_MM256_EXTRACT_EPI64": "#define AWS_HAVE_MM256_EXTRACT_EPI64",
    "#cmakedefine AWS_HAVE_CLMUL": "#define AWS_HAVE_CLMUL",
    "#cmakedefine AWS_HAVE_ARM32_CRC": "",
    "#cmakedefine AWS_HAVE_ARMv8_1": "",
    "#cmakedefine AWS_ARCH_ARM64": "",
    "#cmakedefine AWS_ARCH_INTEL": "#define AWS_ARCH_INTEL",
    "#cmakedefine AWS_ARCH_INTEL_X64": "#define AWS_ARCH_INTEL_X64",
    "#cmakedefine AWS_USE_CPU_EXTENSIONS": "#define AWS_USE_CPU_EXTENSIONS",
}
```

But we could probably just write the config file directly, as we do in [se_curl](https://github.com/google/tensorstore/blob/master/third_party/se_curl/curl.BUILD.bazel)

Or we could specialize the individual cmake lines like we do in [jpeg](https://github.com/google/tensorstore/blob/master/third_party/jpeg/jpeg.BUILD.bazel)

Or maybe we pre-generate files, like we do in [cares](https://github.com/google/tensorstore/tree/master/third_party/com_github_cares_cares), although the per-system #cmakedefine exists there as well.

laramiel:
... it might be nice to contribute aws rules to something like the bazel-central repository: https://github.com/bazelbuild/bazel-central-registry/tree/main/modules

But that happened after we started, so we haven't done that with our other modules, and we use the old-style workspace approach.

sjperkins:
The build now includes more recent version of the AWS C, CRT and C++ SDKs.

FWIW the `s3_sdk` localstack test cases pass locally on linux, but this PR hasn't been tested on Windows or Mac. Unfortunately, I may not have extra bandwidth (or easy access to hardware) for these environments.  In these cases, best-guesses have been applied in the BUILD scripts, but require further scrutiny. Might it be possible to try this out on your end and iron out the differences in review?

There's also probably scope to use the `aws-cpp-sdk-s3-crt` for improved performance, instead of the original `aws-cpp-sdk-s3`.

- https://aws.amazon.com/blogs/storage/improving-amazon-s3-throughput-for-the-aws-cli-and-boto3-with-the-aws-common-runtime/

laramiel:
I've been getting internal imports of these aws c libraries; I wonder if we should not include the aws_cpp_sdk libraries at all, and just limit the use to the aws_crt_cpp libraries; it feels like the aws-cpp-sdk-s3-crt library belongs in aws_crt_cpp...

sjperkins:
>  I wonder if we should not include the aws_cpp_sdk libraries at all, and just limit the use to the aws_crt_cpp libraries; it feels like the aws-cpp-sdk-s3-crt library belongs in aws_crt_cpp...

I think `aws-cpp-sdk-s3-crt` still has a dependency on `aws-cpp-sdk-core` so it's still seems to be part of the AWS C++ SDK library,

https://github.com/aws/aws-sdk-cpp/blob/77b525c975bb2e9840661d4f5beaf5171c8c3673/generated/src/aws-cpp-sdk-s3-crt/include/aws/s3-crt/S3CrtClient.h#L7-L22

and `aws-cpp-sdk-core` has a dependency on the `aws-cpp-crt`

https://github.com/aws/aws-sdk-cpp/blob/77b525c975bb2e9840661d4f5beaf5171c8c3673/src/aws-cpp-sdk-core/include/aws/core/Aws.h#L16-L17

> Mostly, I think, we want the aws_c_auth capabilities to sign requests.

Without having looked at the internals of `aws_c_auth`, this seems plausible.

The benefit of using `aws-cpp-sdk-s3-crt` would be the depending on reference implementations of:

- retry logic
- xml response handling
- endpoint setup

which may be preferable to maintaining tensorstore implementations of the above.

sjperkins:
I'm travelling in July but I'll try find space to swap in the S3CrtClient into the test cases.

laramiel:
I'm currently getting these aws libraries up to aws-crt-cpp imported into our internal build system. I think that I'd like to start with the aws-crt-cpp / aws_c_s3 libraries. 

The aws-cpp-sdk-s3-crt library seems to be built on the aws-c-s3 + aws-crt-cpp along with autogenerated mechanisms to parse the xml responses (in the model directory).  However it's likely that the raw aws-c-s3 libraries will have most of the official retry, etc. done without the xml response parsing, and I'd like to target that as a first attempt.

To that end I may import the third_party bits that you have here.  If you had time to separate that into a distinct pull request it would be great.

sjperkins:
> I'm currently getting these aws libraries up to aws-crt-cpp imported into our internal build system. I think that I'd like to start with the aws-crt-cpp / aws_c_s3 libraries.
> 
> The aws-cpp-sdk-s3-crt library seems to be built on the aws-c-s3 + aws-crt-cpp along with autogenerated mechanisms to parse the xml responses (in the model directory). However it's likely that the raw aws-c-s3 libraries will have most of the official retry, etc. done without the xml response parsing, and I'd like to target that as a first attempt.
> 
> To that end I may import the third_party bits that you have here. If you had time to separate that into a distinct pull request it would be great.

Sure. I'll aim to make that available by the end of this week,

sjperkins:
> ... I actually plan on getting the third_party working since it needs to work for windows / mac.
> So just stay tuned.

OK, just to overcommunicate: I take this to mean to hold off on the aws third_party PR. Let me know if I'm misinterpreting, otherwise I'll be waiting on your changes :-)

laramiel:
Yes; I published my WIP here for now: https://github.com/google/tensorstore/commit/22a7846d59ce0ed0795a9c2dc5d971694c1d8f68

I think that it's pretty close.

sjperkins:
Thanks, that build looks more comprehensive. Regarding the `aws-c-s3` approach, I could try a PR on this when I return at the start of August. If that timeline is too far out for any deliverables, please don't wait on me :-)

laramiel:
I don't have a timeline for updating s3; I work on it in between other things, somewhat like you do.

Anyway, some pointers / ideas.

The aws-c-s3 github repository has an s3 sample which covers most of what we need to do with s3: reading and listing.
https://github.com/awslabs/aws-c-s3/blob/main/samples/s3/

sjperkins:
>  I could try a PR on this when I return at the start of August. 

Apologies, I haven't had time to get to this due to a bout of illness and conferences. Hoping to pick it up next week.

sjperkins:
S3 now supports conditional writes, although other implementations may not:

https://aws.amazon.com/about-aws/whats-new/2024/08/amazon-s3-conditional-writes/

sjperkins:
> Apologies, I haven't had time to get to this due to time off sick and conferences. Hoping to pick it up next week.

I need to get some releases out. I'm still interested, but I'm not sure when I'll have space in the near future.

Eamon2009 (approved):

Issue #137, opened 2024-02-16
Suppiort of large files for grpc_kvstore


zlobober:
This issue is related to grpc_kvstore driver, which is currently not available for using with vanilla tensorstore, but was discussed in the issue https://github.com/google/tensorstore/pull/134 and is going to become public.

ML teams in my company are using a privately patched version of tensorstore with grpc_kvstore driver, whose backend is implemented by YTsaurus. Current grpc protocol imposes the following problem: both Read and Write requests are one-shot, which limits the length of a value by 2 GiB, which is a fundamental upper limit of Protobuf message size. Also, it is not a good idea in general to write read or write large blobs within a single request, because it is not a fault-tolerant solution.

Our proposal is to make Read and Write requests respectively server-side and client-side streaming methods, limiting the size of a single message with something reasonable like 32 MiB. It seems that this change may be done in a backward-incompatible manner, as grpc_kvstore interface is not public and stable yet.

If you are ok with this proposal, we would be glad to bring a PR implemeting this idea.

laramiel:
I concur that the read and write need to be streaming gRPC apis; the gcs_grpc driver is likely a reasonable model for the kvstore driver code.

Note that default gRPC server GRPC_ARG_MAX_RECEIVE_MESSAGE_LENGTH is 4MB; the chunk size should be configurable.

PR #152, opened 2024-04-08
Support streaming in tsgrpc kvstore driver.

zlobober:
This commit changes tsgrpc protocol in backward-incompatible manner, addressing issue #137.
After this change, it becomes possible to read and write values that exceed the theoretic maximum
of 2 GiB per proto message.

### DIFF
--- a/tensorstore/kvstore/tsgrpc/kvstore_server.cc
+++ b/tensorstore/kvstore/tsgrpc/kvstore_server.cc
@@ -72,7 +72,8 @@ using ::grpc::CallbackServerContext;
 using ::tensorstore::kvstore::ListEntry;
 using ::tensorstore_grpc::EncodeGenerationAndTimestamp;
 using ::tensorstore_grpc::Handler;
-using ::tensorstore_grpc::StreamHandler;
+using ::tensorstore_grpc::ServerStreamHandler;
+using ::tensorstore_grpc::ClientStreamHandler;
 using ::tensorstore_grpc::kvstore::DeleteRequest;
 using ::tensorstore_grpc::kvstore::DeleteResponse;
 using ::tensorstore_grpc::kvstore::ListRequest;
@@ -102,13 +103,14 @@ auto& list_metric = internal_metrics::Counter<int64_t>::New(
 
 ABSL_CONST_INIT internal_log::VerboseFlag verbose_logging("tsgrpc_kvstore");
 
-class ReadHandler final : public Handler<ReadRequest, ReadResponse> {
-  using Base = Handler<ReadRequest, ReadResponse>;
+class ReadHandler final : public ServerStreamHandler<ReadRequest, ReadResponse> {
+  using Base = ServerStreamHandler<ReadRequest, ReadResponse>;
 
  public:
   ReadHandler(CallbackServerContext* grpc_context, const Request* request,
-              Response* response, KvStore kvstore)
-      : Base(grpc_context, request, response), kvstore_(std::move(kvstore)) {}
+              KvStore kvstore, size_t max_sent_part_bytes)
+      : Base(grpc_context, request), kvstore_(std::move(kvstore)),
+        max_sent_part_bytes_(max_sent_part_bytes) {}
 
   void Run() {
     ABSL_LOG_IF(INFO, verbose_logging)
@@ -135,14 +137,20 @@ class ReadHandler final : public Handler<ReadRequest, ReadResponse> {
 
     internal::IntrusivePtr<ReadHandler> self{this};
     future_ =
-        PromiseFuturePair<void>::Link(
-            [self = std::move(self)](tensorstore::Promise<void> promise,
-                                     auto read_result) {
-              if (!promise.result_needed()) return;
-              promise.SetResult(self->HandleResult(read_result.result()));
-            },
-            tensorstore::kvstore::Read(kvstore_, request()->key(), options))
-            .future;
+      PromiseFuturePair<void>::Link(
+          [self = std::move(self)](tensorstore::Promise<void> promise,
+                                            auto read_result) {
+            if (!promise.result_needed()) {
+              self->Finish(::grpc::Status(::grpc::StatusCode::CANCELLED, ""));
+              return;
+            }
+            // Actual value set to the promise does not matter, as the promise is only used for tracking completion or
+            // cancellation.
+            promise.SetResult(absl::OkStatus());
+            self->HandleResult(read_result.result());
+          },
+          tensorstore::kvstore::Read(kvstore_, request()->key(), options))
+          .future;
   }
 
   void OnCancel() final {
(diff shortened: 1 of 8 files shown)


zlobober:
@laramiel, would you mind having a look?

laramiel:
Looking; generally looks reasonable, will try to give it more of a review tomorrow, or early next week.

tensorstore/kvstore/tsgrpc/tsgrpc.cc
```diff
@@ -137,7 +138,11 @@ struct TsGrpcKeyValueStoreSpecData {
       jb::Member(
           DataCopyConcurrencyResource::id,
           jb::Projection<
-              &TsGrpcKeyValueStoreSpecData::data_copy_concurrency>()) /**/
+              &TsGrpcKeyValueStoreSpecData::data_copy_concurrency>()),
+      jb::Member("max_sent_part_bytes",
```
laramiel, line 142: I think that we just use something like this:

constexpr size_t kMaxPayloadSize = 2 * 1024 * 1024;  // 2mb, gRPC allows 4mb.

tensorstore/kvstore/tsgrpc/tsgrpc.cc
```diff
@@ -215,77 +229,169 @@ struct ReadTask : public internal::AtomicReferenceCount<ReadTask> {
...
 /// Implements `TsGrpcKeyValueStore::Write`.
-struct WriteTask : public internal::AtomicReferenceCount<WriteTask> {
+struct WriteTask : public internal::AtomicReferenceCount<WriteTask>, public grpc::ClientWriteReactor<WriteRequest> {
   internal::IntrusivePtr<TsGrpcKeyValueStore> driver;
   grpc::ClientContext context;
   WriteRequest request;
   WriteResponse response;
 
+  absl::Cord value;
+  size_t value_offset = 0;
```
laramiel, line 310: #include <stddef.h>

tensorstore/kvstore/tsgrpc/tsgrpc.cc
```diff
@@ -215,77 +229,169 @@ struct ReadTask : public internal::AtomicReferenceCount<ReadTask> {
...
         << "WriteTask::Ready " << ConciseDebugString(response) << " " << status;
     TENSORSTORE_RETURN_IF_ERROR(status);
     TENSORSTORE_RETURN_IF_ERROR(GetMessageStatus(response));
     return DecodeGenerationAndTimestamp(response);
   }
+
+  void SetNextPart() {
+    auto next_part = value.Subcord(value_offset, driver->GetMaxSentPartBytes());
+    request.set_value_part(std::move(next_part));
+    value_offset = std::min(value.size(), value_offset + next_part.size());
```
laramiel, line 393: #include <algorithm>

tensorstore/kvstore/tsgrpc/tsgrpc.cc
laramiel: ... which means that there should be a separate test which uses a 4mb buffer, or so.

tensorstore/kvstore/tsgrpc/kvstore_server_test.cc
```diff
@@ -91,6 +94,20 @@ TEST_F(KvStoreTest, Basic) {
   tensorstore::internal::TestKeyValueReadWriteOps(store);
 }
 
+TEST_F(KvStoreTest, Multipart) {
+  auto context = tensorstore::Context::Default();
```
laramiel, line 98: Use something like this:

```
  // uses 2mb chunks; this will force at least two writes.
  tensorstore::internal::FlatCordBuilder cord_builder(16 + 2 * 1024 * 1024);
  memset(cord_builder.data(), 0x37, cord_builder.size());
  absl::Cord data = std::move(cord_builder).Build();
  data.Append("abcd");  // second chunk.
```

With these includes:

```
#include <cstring>
#include <utility>
#include "third_party/tensorstore/internal/flat_cord_builder.h"
```

laramiel: Likewise in tsgrpc_test.cc

tensorstore/kvstore/tsgrpc/tsgrpc_test.cc
```diff
@@ -77,16 +77,28 @@ class TsGrpcMockTest : public testing::Test {
...
                              {"driver", "tsgrpc_kvstore"},
                              {"address", mock_service_.server_address()},
+                             {"max_sent_part_bytes", max_sent_part_bytes},
                          })
         .value();
   }
 
   MockKvStoreService& mock() { return *mock_service_.service(); }
 
+  void ExpectWriteRequestsAre(grpc::ServerReader<WriteRequest>* req,
```
laramiel, line 91: Avoid using EXPECT in nested functions; instead consider adding a matcher.

https://google.github.io/googletest/reference/matchers.html#defining-matchers

tensorstore/kvstore/tsgrpc/kvstore_server.cc
```diff
@@ -151,49 +159,101 @@ class ReadHandler final : public Handler<ReadRequest, ReadResponse> {
...
+      } else {
+        // Perform actual writing.
+        internal::IntrusivePtr<WriteHandler> self{this};
+        future_ =
+          PromiseFuturePair<void>::Link(
+              [self = std::move(self)](Promise<void> promise, auto write_result) {
+                if (!promise.result_needed()) return;
+                // Actual value set to the promise does not matter, as the promise is only used for tracking completion
+                // or cancellation.
+                promise.SetResult(absl::OkStatus());
```
laramiel, line 234: I think that this should be more like:

promise.SetResult(self->HandleResult(write_result.result()));

Or at least, the HandleResult() needs the status response handled correctly.

tensorstore/kvstore/tsgrpc/kvstore_server.cc
```diff
@@ -440,32 +507,37 @@ TENSORSTORE_DEFINE_JSON_DEFAULT_BINDER(
                }),
                jb::Member("bind_addresses",
                           jb::Projection<&KvStoreServer::Spec::bind_addresses>(
-                              jb::DefaultInitializedValue()))));
+                              jb::DefaultInitializedValue())),
+               jb::Member("max_sent_part_bytes",
```
laramiel, line 511: Hm, if we want to support a parameter like this, then it needs to be validated to be within the range [1, 4mb - ~16), otherwise it will cause gRPC issues.

zlobober: I actually considered such naming but decided that it would be sort of misleading, as this parameter limits the size of the "value_part" field of the protobuf and not the total size of the message (which is typically slightly larger due to the rest of the fields).

Anyway, this should not be a big deal, so let me change the naming as you propose.

laramiel:
If you sync this, then I can look at applying it.

zlobober:
Was on conference/sick leave, sorry for long time without response.
Actually I am working on fixing everything right now, will bring an update later today/tomorrow.
Thanks!

tensorstore/kvstore/tsgrpc/kvstore_server_test.cc
zlobober: What's the idea behind this change? Would you like to (implicitly) fix the fact that default value of `max_sent_part_bytes` (or `max_message_bytes` as you proposed to rename) equals to 2 mib? If yes, I can alternatively introduce a separate test that ensures that default part size is 2 mib without changing much the existing tests; to me it looks like they do their job and cover all "chunking" logic.

Also, IMHO using long strings in tsgrpc_test.cc would be less elegant, as the rest of tests constructs proto messages using ParseTextProtoOrDie, which makes understanding the communication quite easy. 

But if you insist, I'll rewrite all unittests using 2 mib chunks.

Also, why do you suggest `16 + 2 * 1024 * 1024`? The first chunk would in fact be the first 2 mib, and the second would consist of 16 times (char)0x37 and abcd in your example, and not of "abcd" solely as your comment suggests.

laramiel:
No longer required

PR #154, opened 2024-04-18
Remove tsgrpc_kvstore address log

Krock21:
Hi!

We are working on integrating tensorstore with [YTsaurus](https://github.com/ytsaurus/ytsaurus) as an underlying storage.

We were able to do some PoC and received some feedback from our customers already:
`tsgrpc_kvstore address=` log seems not very useful and makes a lot of noise in some cases

We believe that most users would prefer not to have this log line

Cheers,
Aleksandr.


laramiel:
I adjusted the logging to only happen when verbose is enabled via https://github.com/google/tensorstore/commit/67ba2cda1c6fdfc62d130320eb2d43f866b63885

PR #187, opened 2024-08-12
ec2_credential_provider supports IMDSv2

jonathan-marcus:
EC2 credentials provider currently supports three methods: env vars, file, and IMDS. It only queries v1, however, so if only v2 is enabled then it gets a 405 exception and fails.

This change catches that 405 (as well as a 401 that I believe could exist also) and tries a PUT to hit IMDSv2 as a fallback.


laramiel:
This seems reasonable; I'll try and get it merged.

Note that sjperkins@ is working on migrating some of this to the s3 C libraries; if you wanted to work on that you could probably collaborate somehow.

PR #229, opened 2025-04-14
Add codespell support (config, workflow to detect/not fix) and make it fix a typo

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

I eyeball the hits, there is even a fix to a URL to become proper; the more eyes the merrier.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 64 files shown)


laramiel:
Thanks for the cleanups. I can apply the spelling fixes soon, but will hold off on the automation since we need to figure out how to run it internally as well.

yarikoptic:
locally -- just `pip install codespell` (or via anything else) and call `codespell`.  Some folks add to pre-commit when they use that, e.g. like we in https://github.com/dandi/dandi-cli/blob/master/.pre-commit-config.yaml#L21

laramiel:
Do you know if there are existing Bazel codespell macros? It would make integration pretty easy if we could run it as a bazel test.

yarikoptic:
sorry , not familiar with bazel. Nothing 100% obvious with quick google search came up.

PR #230, opened 2025-04-17
Add TIFF Key-Value Store Implementation (take 2)

hsidky:
This PR adds a new (read only) key-value store driver for accessing tiles and strips within TIFF image files. This is my second attempt after finding the time to work on this given previous feedback on #138. I drew heavily upon the zip kvstore implementation and your prior feedback (thanks!), so I hope that things are more faithful to Tensorstore patterns.

The basic idea is this kvstore sits on top of a base kvstore and maps IFDs/tiles/stripes to the following key format: `tile/<id>/<row>/<col>`. 

My plan is to get feedback on whether this something that is suitable to be merged. If I receive a positive response, I work on any feedback I receive in parallel with building out the tiff driver that will pair with this kvstore.

Looking forward to your feedback on this implementation!

### DIFF
--- a/tensorstore/kvstore/tiff/tiff_details.h
+++ b/tensorstore/kvstore/tiff/tiff_details.h
@@ -75,6 +75,9 @@ struct IfdEntry {
   uint64_t count;
   uint64_t value_or_offset;  // For values that fit in 4/8 bytes, this is the value
                             // Otherwise, this is an offset to the data
+  
+  // Flag to indicate if this entry references an external array
+  bool is_external_array = false;
 };
 
 // Represents a TIFF Image File Directory (IFD)
@@ -118,6 +121,21 @@ absl::Status ParseTiffDirectory(
 absl::Status ParseImageDirectory(
     const std::vector<IfdEntry>& entries,
     ImageDirectory& out);
+
+// Parse an external array from a reader
+absl::Status ParseExternalArray(
+    riegeli::Reader& reader,
+    Endian endian,
+    uint64_t offset,
+    uint64_t count,
+    TiffDataType data_type,
+    std::vector<uint64_t>& out);
+
+// Determine if an IFD entry represents an external array based on type and count
+bool IsExternalArray(TiffDataType type, uint64_t count);
+
+// Get the size in bytes for a given TIFF data type
+size_t GetTiffDataTypeSize(TiffDataType type);
 
 }  // namespace internal_tiff_kvstore
 }  // namespace tensorstore
--- a/tensorstore/kvstore/tiff/tiff_details.h
+++ b/tensorstore/kvstore/tiff/tiff_details.h
@@ -34,88 +34,88 @@ enum class Endian {
 };
 
 enum Tag : uint16_t {
-  kImageWidth        = 256,
-  kImageLength       = 257,
-  kBitsPerSample     = 258,
-  kCompression       = 259,
-  kPhotometric       = 262,
-  kSamplesPerPixel   = 277,
-  kRowsPerStrip      = 278,
-  kStripOffsets      = 273,
-  kStripByteCounts   = 279,
-  kPlanarConfig      = 284,
-  kTileWidth         = 322,
-  kTileLength        = 323,
-  kTileOffsets       = 324,
-  kTileByteCounts    = 325,
-  kSampleFormat      = 339,
+  kImageWidth = 256,
+  kImageLength = 257,
+  kBitsPerSample = 258,
+  kCompression = 259,
+  kPhotometric = 262,
+  kSamplesPerPixel = 277,
+  kRowsPerStrip = 278,
+  kStripOffsets = 273,
+  kStripByteCounts = 279,
+  kPlanarConfig = 284,
+  kTileWidth = 322,
+  kTileLength = 323,
+  kTileOffsets = 324,
+  kTileByteCounts = 325,
+  kSampleFormat = 339,
 };
 
 // Common compression types
 enum class CompressionType : uint16_t {
-  kNone       = 1,
+  kNone = 1,
   kCCITTGroup3 = 2,
   kCCITTGroup4 = 3,
-  kLZW        = 5,
-  kJPEG       = 6,
-  kDeflate    = 8,
-  kPackBits   = 32773,
+  kLZW = 5,
+  kJPEG = 6,
+  kDeflate = 8,
+  kPackBits = 32773,
 };
 
 // Photometric interpretations
 enum class PhotometricType : uint16_t {
   kWhiteIsZero = 0,
   kBlackIsZero = 1,
-  kRGB        = 2,
-  kPalette    = 3,
+  kRGB = 2,
+  kPalette = 3,
   kTransparencyMask = 4,
-  kCMYK       = 5,
-  kYCbCr      = 6,
-  kCIELab     = 8,
+  kCMYK = 5,
+  kYCbCr = 6,
+  kCIELab = 8,
 };
 
 // Planar configurations
 enum class PlanarConfigType : uint16_t {
-  kChunky     = 1,  // RGBRGBRGB...
-  kPlanar     = 2,  // RRR...GGG...BBB...
+  kChunky = 1,  // RGBRGBRGB...
+  kPlanar = 2,  // RRR...GGG...BBB...
 };
 
 // Sample formats
 enum class SampleFormatType : uint16_t {
-  kUnsignedInteger  = 1,
-  kSignedInteger    = 2,
-  kIEEEFloat        = 3,
-  kUndefined        = 4,
+  kUnsignedInteger = 1,
+  kSignedInteger = 2,
+  kIEEEFloat = 3,
+  kUndefined = 4,
 };
 
 // TIFF data types
 enum class TiffDataType : uint16_t {
-  kByte = 1,      // 8-bit unsigned integer
-  kAscii = 2,     // 8-bit bytes with last byte null
-  kShort = 3,     // 16-bit unsigned integer
-  kLong = 4,      // 32-bit unsigned integer
-  kRational = 5,  // Two 32-bit unsigned integers
-  kSbyte = 6,     // 8-bit signed integer
-  kUndefined = 7, // 8-bit byte
-  kSshort = 8,    // 16-bit signed integer
-  kSlong = 9,     // 32-bit signed integer
-  kSrational = 10,// Two 32-bit signed integers
-  kFloat = 11,    // 32-bit IEEE floating point
-  kDouble = 12,   // 64-bit IEEE floating point
-  kIfd = 13,      // 32-bit unsigned integer (offset)
-  kLong8 = 16,    // BigTIFF 64-bit unsigned integer
-  kSlong8 = 17,   // BigTIFF 64-bit signed integer
-  kIfd8 = 18,     // BigTIFF 64-bit unsigned integer (offset)
+  kByte = 1,        // 8-bit unsigned integer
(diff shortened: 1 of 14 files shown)


jbms:
This is great --- I haven't had a chance to review your implementation in detail, but at a high level this is exactly the right implementation strategy.

Do you have a reason to use this tiff kvstore driver on its own?  Otherwise I'd be inclined to eliminate the JSON registration for it, such that it can only be created internally (by the to-be-written Tiff TensorStore driver).

jbms:
Regarding logging: what have you tried?  Are you testing with bazel?  With bazel you can do `--test_env=TENSORSTORE_VERBOSE_LOGGING=tiff`

hsidky:
> This is great --- I haven't had a chance to review your implementation in detail, but at a high level this is exactly the right implementation strategy.
> 
> Do you have a reason to use this tiff kvstore driver on its own? Otherwise I'd be inclined to eliminate the JSON registration for it, such that it can only be created internally (by the to-be-written Tiff TensorStore driver).

Thanks for that feedback. No, I don't think there is any reason to use the tiff kvstore on its own. I can eliminate the JSON registration. 

> Regarding logging: what have you tried? Are you testing with bazel? With bazel you can do `--test_env=TENSORSTORE_VERBOSE_LOGGING=tiff`

Yes, I'm using bazel though I'm not terribly familiar with it. That works! I had previously tried setting the environment variable as that worked for non-subsystem logging in tests, but it wasn't working for conditional logging. Thank you. 

I'm going to start chipping away at the tiff driver as I await detailed feedback on this kvstore.

laramiel:
The closest existing similar code is probably neuroglancer_uint64_sharded.cc

The neuroglancer implementation uses several kvstores; you could do something similar where there is a TiffMetadataKvStore which returns metadata entries for a tiff.  This could use the same underlying kvstore for access to the tiff file.

jbms:
You have made some good observations --- kvs_backed_chunk_driver indeed currently assumes a zarr-like structure.

Ultimately it could be nice if we refactored things so that some of the logic of kvs_backed_chunk_driver could be shared, but doing that refactor may be tricky and I'm not sure how helpful it will be given the nature of the tiff driver.

Instead I'd suggest:

1. Implement a TiffChunkCache that inherits from KvsBackedChunkCache and uses your tiff chunk kvstore to read chunks.
2. Implement an `internal::Driver` (using RegisteredDriver) that contains a `CachePtr<TiffChunkCache>` and also e.g. a PinnedCacheEntry pointer to the relevant entry in your tiff dir cache.  This will have duplicate a small amount of logic from kvs_backed_chunk_driver, but that will also help highlight where any refactoring could be beneficial.

I think your existing tiff dir cache is sufficient and you don't need any additional metadata cache.

O

hsidky:
Thank you both for your feedback. I'll take a stab at this and report back with any questions.

hsidky:
Hello again. I've managed to develop a driver per your suggestions and I believe I have something that's a pretty solid foundation with a reasonable amount of support. There are some outstanding items + polish but I think it makes sense for me to pause here for feedback in case major changes are required and to understand the likelihood of being able to get this merged first. 

A few notes on the driver implementation:
- I haven't worked on transaction support as I haven't fully wrapped my head around the nuts and bolts of how that works yet. Is this a requirement for a merge? I saw, for example, that the single image driver does not support transactions, so I assume not. 
- Same as above but for storage stats. 
- I did not commit the test file binaries (generated by the generate.py file). I'll wait until you give me clearance. 
- I waffled a bit on how much to build out the metadata property. I ended up putting options under a separate tiff property as they mainly correspond to the _interpretation_ of the data. I can revisit if you have strong opinions. 

I think that's it for now. Looking forward to any feedback!

tensorstore/driver/tiff/schema.yml
```diff
@@ -0,0 +1,166 @@
...
+              type: array
+              items:
+                type: integer
+                minimum: 1
+              title: Sizes of stacked dimensions.
+              description: |
+                Optional. Must match the length of `dimensions`. Required if `dimensions` 
+                has more than one entry. If only one dimension is specified, `ifd_count` can 
+                be used instead. If both are specified, their product must match `ifd_count`.
+            ifd_count:
```
jbms, line 78: Why does this need to be specified --- can't we determine the number of IFDs from the file?

tensorstore/driver/tiff/schema.yml
```diff
@@ -0,0 +1,166 @@
...
+              type: integer
+              minimum: 1
+              title: Total number of IFDs in the stack.
+              description: |
+                Optional. Specifies the total number of IFDs involved in the stack.
+                Required if `dimension_sizes` is not specified for a single stack dimension.
+                If specified along with `dimension_sizes`, their product must match `ifd_count`.
```
jbms, line 85: I think `dimensions`, `dimension_sizes` and `ifd_sequence_order` can be removed as options.  Instead, there should be a single unnamed dimension as the "stack" dimension.  Then a separate `reshape` adapter TensorStore driver (not yet implemented, but on the todo list can be responsible for doing the equivalent of those options.

jbms: (Instead of an unlabeled dimension, the stack dimension could also be assigned some default label like `stack`)

tensorstore/driver/tiff/schema.yml
```diff
@@ -0,0 +1,166 @@
...
+        they must match the corresponding properties inferred from the TIFF tags
+        and interpretation options.
+      allOf:
+      - type: object
+        properties:
+          dtype:
+            $ref: DataType
+            title: Data type constraint.
+            description: Constrains the expected data type of the TIFF dataset.
+          shape:
```
jbms, line 28: This can likewise be excluded since it can be specified via the `schema`.

tensorstore/driver/tiff/schema.yml
hsidky: I suppose it doesn't. I initially was thinking about wanting to use only a subset of IFDs, but then really `ifd_count` would be one of many parameters that a user would need to specify in order to provide them with the kind of flexibility to do that properly. I will remove.

tensorstore/driver/tiff/schema.yml
hsidky: I think the issue with that, if I understand enough about how tensorstore works, is that this will become an problem when we are pulling that metadata from the file itself (e.g. OME-TIFF), versus having the user specify that information (which they may not know since it's embedded in the image) using a `reshape` adapter. 

The reason I added these in here was to build out the necessary functionality and options such that I can later on just read the XML data from the appropriate TIFF tag and assign it to existing parameters. I did not add that in yet because I wanted to get the base, non-OME-specific, TIFF reader cleared first. Then adding in the XML parsing and populating these fields becomes a trivial exercise. 

If you feel differently, or would still prefer that I just assign a single stacking dimension, I can do that. But (a) I think that means a longer runway for OME-TIFF support and (b) I don't know then how the XML metadata would work its way up to the reshape adapter. 

Please advise.

tensorstore/driver/tiff/schema.yml
```diff
@@ -0,0 +1,166 @@
...
+                type: string
+              uniqueItems: true
+              title: IFD sequence iteration order.
+              description: |
+                Optional. Specifies the order of stacked dimensions within the flat IFD sequence. 
+                Must be a permutation of `dimensions`. Defaults to the order in `dimensions`, with 
+                the last dimension varying fastest.
+          required:
+          - dimensions
+        sample_dimension_label:
```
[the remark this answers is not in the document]
hsidky: The reason I added this is because OME XML has dimension labels in its metadata (XYZCT). I referenced the n5 driver a lot in my development and I saw that it has an `axes` member for dimension labels in N5Metadata. As with my previous comment, right now, they are manually populated, but when the OME XML is parsed, I will populate this. 

I realize I am in some ways optimizing for OME XML while trying to make a generic solution for TIFF images. The way I had envisioned it was that these are generic options for anyone to use to craft some interpretation of their TIFF file, though I can see how it can look a bit odd. Also happy to follow your guidance on this. Just explaining my rationale.

tensorstore/driver/tiff/schema.yml
```diff
@@ -0,0 +1,166 @@
...
+      title: TIFF dataset metadata constraints.
+      description: |
+        Specifies constraints that are validated against the metadata read
+        from the existing TIFF file. All members are optional. If specified,
+        they must match the corresponding properties inferred from the TIFF tags
+        and interpretation options.
+      allOf:
+      - type: object
+        properties:
+          dtype:
```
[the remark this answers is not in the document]
hsidky: For the metadata property (which is just dtype and shape below), I think I ultimately got confused. I was trying to square three things: (a) the concept of metadata constraints that I saw in n5, (b) come up with a generic TIFF solution, and (c) preempt support for OME-XML. The properties that I put inside the `tiff` object  - if we set aside OME XML - are technically not constraints. They specify different ways of _interpreting_ the same underlying data (e.g. whether IFDs are stacked as a dimension or treated as separate images is in the mind of the user, not something the TIFF file is stipulating). It's only with OME-XML that it becomes true metadata. 

I used the metadata object for properties that are not subject to interpretation, like the dtype, and shape (at least partially).  I realize this is quite messy. I could perhaps move the `tiff` object into `metadata`, and then it functions the same way it does with n5: all members are optional, and they are populated if found (e.g. OME XML).  But it will also serve one additional function: which is to specify the _interpretation_ of the underlying data if no OME metadata is present. This is probably cleaner from a schema perspective, but it does mix up two distinct responsibilities. 

Let me know what you prefer.

tensorstore/driver/tiff/schema.yml
hsidky: Actually, I double checked and I see now that IFDs may be ignored in OME XML so I will I need to add full specification functionality rather than remove this. https://docs.openmicroscopy.org/ome-model/5.6.3/ome-tiff/specification.html

tensorstore/driver/tiff/compressor_registry.h
```diff
@@ -0,0 +1,46 @@
...
+#include "tensorstore/kvstore/tiff/tiff_details.h"
+
+namespace tensorstore {
+namespace internal_tiff {
+
+// Returns the global registry instance for TIFF compressors.
+// This registry maps string IDs (like "lzw", "deflate") to factories/binders
+// capable of creating JsonSpecifiedCompressor instances.
+internal::JsonSpecifiedCompressor::Registry& GetTiffCompressorRegistry();
+
```
laramiel, line 32: It seems unlikely that we need to register tiff compressor to start with, also the tiff tag value could map directly to a compressor instance (via an enum, or similar).

tensorstore/kvstore/tiff/tiff_details.h
```diff
@@ -0,0 +1,202 @@
...
+  kPlanarConfig = 284,
+  kTileWidth = 322,
+  kTileLength = 323,
+  kTileOffsets = 324,
+  kTileByteCounts = 325,
+  kSampleFormat = 339,
+};
+
+// Common compression types
+enum class CompressionType : uint16_t {
```
laramiel, line 55: I'd probably find the hex values easier to track back to the spec.

tensorstore/kvstore/tiff/tiff_details.h
```diff
@@ -0,0 +1,202 @@
...
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "absl/status/status.h"
+#include "riegeli/bytes/reader.h"
+
+namespace tensorstore {
+namespace internal_tiff_kvstore {
+
```
laramiel, line 30: Add a reference to the spec sections here as comments.

PR #264, opened 2025-12-04
Add support for structured dtypes to `zarr3` driver, open zarr 2 and 3 structs as void

BrianMichell:
Resolves #241 

This PR implements support for structured data in the `zarr3` driver as well as the ability to interact with the full array as raw bytes to avoid needing to manage multiple Stores in client code. ~~I am also happy to revisit the question of #125 for a more general Zarr support for this behavior.~~ I have also implemented the `open_as_void` feature in the `zarr` driver for feature parity between the formats.

Below is an example workflow that this usecase would support.

The Python snippet will generate a structured array of several different elements and types. A real world example of this is creating an array to store the [SEG-Y 240-byte Trace Headers](https://seg.org/wp-content/uploads/2025/11/seg_y_rev1.pdf#page=15).

<details>
    <summary>Click here to expand Python example.</summary>

```python
import numpy as np
import zarr

store = "foo.zarr"

np_dtype = np.dtype(
    [
        ("field_1", "<i4"),
        ("field_2", ">i4"),
        # More theoretical fields here...
        ("field_n", "<f4"),
    ]
)

z = zarr.create_array(
    store=store,
    shape=(128, 128),
    dtype=np_dtype,
    chunks=(32, 32),
)

arr = np.zeros((128, 128), dtype=np_dtype)

f1 = np.arange(128, dtype="<i4")
# NOTE: We populate f2 as little endian for the demonstration
f2 = np.arange(128, dtype="<i4")
fn = np.arange(128, dtype="<f4") / 10

arr["field_1"][:] = f1[:, None]
arr["field_2"][:] = f2[:, None]
arr["field_n"][:] = fn[:, None]

z[:] = arr
```
</details>

The usecase continues with an HPC C++ application consuming either a single field from the structured dtype, or the entirety of the dtype to perform its intended workload. To keep this example code simple, I only demonstrate consuming the same field out of the raw data, however this could reasonably be extended to 70+ fields per element if writing a trivial data loader.

<details>
    <summary>Click here to expand C++ example.</summary>

```cpp
// Minimal example demonstrating structured Zarr v3 data reading.
// This example reads the 'foo.zarr' output generated by generate_struct.py
//
// Usage:
//   ./read_foo_zarr --zarr_path=/path/to/foo.zarr
#include <stdint.h>

#include <iostream>
#include <string>
#include <cstring>

#include "absl/flags/flag.h"
#include "absl/flags/parse.h"
#include "absl/status/status.h"
#include <nlohmann/json.hpp>
[...]

PR #272, opened 2026-01-05
Add support for opening structured dtypes as void for `zarr` driver

BrianMichell:
Supersedes #264 

Removes `bool open_as_void = false` default in C++ function declarations and definitions.

Uses a derived DataCache implementation similar to strategy used in `neuroglancer_precomputed` driver for `UnshardedDataCache` and `ShardedDataCache`

### DIFF
--- a/tensorstore/driver/zarr/driver.cc
+++ b/tensorstore/driver/zarr/driver.cc
@@ -410,37 +410,15 @@ std::string DataCache::GetBaseKvstorePath() { return key_prefix_; }
 
 absl::Status VoidDataCache::ValidateMetadataCompatibility(
     const void* existing_metadata_ptr, const void* new_metadata_ptr) {
-  assert(existing_metadata_ptr);
+  // The existing metadata is already void metadata (from cache initialization).
+  // Convert the new metadata to void metadata so both have the same synthesized
+  // void dtype, then use normal validation which compares all fields except
+  // shape (via IsMetadataCompatible).
   assert(new_metadata_ptr);
-  const auto& existing_metadata =
-      *static_cast<const ZarrMetadata*>(existing_metadata_ptr);
   const auto& new_metadata =
       *static_cast<const ZarrMetadata*>(new_metadata_ptr);
-
-  // For void access, we only require that bytes_per_outer_element matches,
-  // since we're treating the data as raw bytes regardless of the actual dtype.
-  // Shape is allowed to differ (handled by base class for resizing).
-  // Other fields like compressor, order, chunks must still match.
-  if (existing_metadata.dtype.bytes_per_outer_element !=
-      new_metadata.dtype.bytes_per_outer_element) {
-    return absl::FailedPreconditionError(tensorstore::StrCat(
-        "Void access metadata bytes_per_outer_element mismatch: existing=",
-        existing_metadata.dtype.bytes_per_outer_element,
-        ", new=", new_metadata.dtype.bytes_per_outer_element));
-  }
-
-  // Check that other critical fields match (same as base, but ignoring dtype)
-  if (existing_metadata.chunks != new_metadata.chunks) {
-    return absl::FailedPreconditionError("Chunk shape mismatch");
-  }
-  if (existing_metadata.order != new_metadata.order) {
-    return absl::FailedPreconditionError("Order mismatch");
-  }
-  if (existing_metadata.compressor != new_metadata.compressor) {
-    return absl::FailedPreconditionError("Compressor mismatch");
-  }
-
-  return absl::OkStatus();
+  return DataCache::ValidateMetadataCompatibility(
+      existing_metadata_ptr, new_metadata.GetVoidMetadata().get());
 }
 
 absl::Status VoidDataCache::GetBoundSpecData(
--- a/tensorstore/driver/zarr/driver_impl.h
+++ b/tensorstore/driver/zarr/driver_impl.h
@@ -166,14 +166,14 @@ class DataCache : public internal_kvs_backed_chunk_driver::DataCache {
 /// The void metadata (created via CreateVoidMetadata) has dtype.fields
 /// containing only the void field, so inherited encode/decode methods
 /// work correctly for raw byte access. GetBoundSpecData is overridden
-/// to set open_as_void=true in the spec, and ValidateMetadataCompatibility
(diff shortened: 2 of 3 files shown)


tensorstore/driver/zarr/driver.cc
```diff
@@ -382,6 +416,178 @@ Result<ChunkLayout> DataCache::GetChunkLayoutFromMetadata(
...
+    if (!base_writer.Close()) return base_writer.status();
+    return encoded;
+  }
+  return uncompressed;
+}
+
+absl::Status VoidDataCache::GetBoundSpecData(
+    internal_kvs_backed_chunk_driver::KvsDriverSpec& spec_base,
+    const void* metadata_ptr, size_t component_index) {
+  auto& spec = static_cast<ZarrDriverSpec&>(spec_base);
```
laramiel: TENSORSTORE_RETURN_IF_ERROR(DataCache::GetBoundSpecData(...))
auto& spec = static_cast<ZarrDriverSpec&>(spec_base);
spec.open_as_void = true;
return absl::OkStatus();

tensorstore/driver/zarr/driver.cc
```diff
@@ -195,6 +182,11 @@ TENSORSTORE_DEFINE_JSON_DEFAULT_BINDER(
                        jb::DefaultValue<jb::kNeverIncludeDefaults>(
                            [](auto* v) { *v = false; }))),
         jb::Initialize([](auto* obj) {
+          // Validate that field and open_as_void are mutually exclusive
```
laramiel: there are helpers for some of this.

jb::Object(
  jb::AtMostOne("field", "open_as_void"),
  ... other jb stuff.

tensorstore/driver/zarr/spec.cc
```diff
@@ -172,7 +173,12 @@ Result<ZarrMetadataPtr> GetNewMetadata(
     // multi-field zarr dtype is desired, it must be specified explicitly.
     TENSORSTORE_ASSIGN_OR_RETURN(
```
laramiel: This seems a little roundabout. I'd need to look at where GetFieldIndex is used, but it seeems like it might be more direct to do something like (spitballing here):

if (open_as_void) {
  metadata->dtype = ... open_as_void dtype with a single dtype.field set ...
} else {
  metadata->dtype = *partial_metadata.dtype;
}

tensorstore/driver/zarr/schema.yml
```diff
@@ -8,15 +8,28 @@ allOf:
...
-      field:
-        oneOf:
-          - type: string
-          - type: "null"
-        title: Name of field to open.
-        description: |
-          Must be specified if the `.metadata.dtype` specified in the array
-          metadata has more than one field.
-        default: null
+    oneOf:
```
jbms: This isn't a valid schema --- but also I don't think we actually want the oneOf constraint since specifying `open_as_void=false` along with non-empty `field` name should be allowed

laramiel: I was thinking about field vs. open_as_void as a variant.

tensorstore/driver/zarr/driver.cc
BrianMichell: So it seems to me that we can resolve this comment based on the conversation from [schema.yml](https://github.com/google/tensorstore/pull/272/changes/BASE..7fb91d751f24a1dd479fd389378f9e6ad33c8112#r2669500148), correct?

BrianMichell:
Sorry for the delay responding to the latest round of feedback. I believe everything raised so far has been addressed.

tensorstore/driver/zarr/spec.h
```diff
@@ -70,9 +70,11 @@ using SelectedField = std::string;
 /// \param partial_metadata Constraints in the form of partial zarr metadata.
 /// \param selected_field The field to which `schema` applies.
 /// \param schema Schema constraints for the `selected_field`.
+/// \param open_as_void If true, opens the array as raw bytes.
```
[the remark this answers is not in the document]
laramiel: Also for GetSpecRankAndFieldInfo

tensorstore/driver/zarr/dtype.h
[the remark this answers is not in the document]
laramiel: And the BUILD deps for the new non-std library includes as well.

tensorstore/driver/zarr/driver.cc
```diff
@@ -496,17 +579,28 @@ class ZarrDriver::OpenState : public ZarrDriver::OpenStateBase {
...
-    const auto& metadata =
+    const auto& original_metadata =
         *static_cast<const ZarrMetadata*>(initializer.metadata.get());
+    auto dim_sep = GetDimensionSeparator(spec().partial_metadata, original_metadata);
+    if (spec().open_as_void) {
+      // Create void metadata from the original. This modifies the dtype to
+      // contain only the void field, allowing standard encode/decode to work.
+      // CreateVoidMetadata uses the same chunks and bytes_per_outer_element as
+      // the original validated metadata, so it should never fail.
+      initializer.metadata = CreateVoidMetadata(original_metadata).value();
```
jbms: Probably would be better to have the original metadata cache the pointer to the void metadata and initialize it on first access.

tensorstore/driver/zarr/driver.cc
```diff
@@ -515,8 +609,16 @@ class ZarrDriver::OpenState : public ZarrDriver::OpenStateBase {
...
+    // component for raw byte access
+    size_t field_index;
+    if (spec().open_as_void) {
+      field_index = 0;
+    } else {
+      TENSORSTORE_ASSIGN_OR_RETURN(
+          field_index,
+          GetFieldIndex(metadata.dtype, spec().selected_field));
+    }
     TENSORSTORE_RETURN_IF_ERROR(
```
jbms: we need to make sure to validate the schema against the void metadata, not the regular metadata.  Note: The partial_metadata validation is still against the regular metadata, though.

laramiel:
I think that this is getting pretty close.  I have a few minor edits in my import which I can just add in.  Getting jeremy to look over it. Thanks.

Edits:
  * some include changes.  The riegeli includes are no longer needed in driver.cc any longer.
  * rename VoidMetadataCache -> LazyVoidMetadata.  Same for VoidFieldCache -> LazyVoidField.
  * convert Lazy... into class with private members & friend to the enclosing struct.

PR #277, opened 2026-01-22
Add batch to VirtualChunked ReadParameters

### LINKED ISSUE #276: Support passing batches to virtual_chunked
- #262 has additional context

I've been using this sort of pattern to access data in a http kvstore

```python
import tensorstore as ts

def virtual_store_factory(endpoint: str, chunk_info: Dict):
  http_spec = { "driver": "http", "base_url": endpoint }
  http_store = ts.KvStore.open(http_spec).result()
  
  def read(domain, array, params):
    key_bits = [f"{o:05}" for o in domain.origin]
    key = f"prefix/{'_'.join(key_bits)}.npy"
    read_result = http_store.read(key).result()
    data = np.frombuffer(read_result.value[header_len:], dtype=dtype)   # Simplification
    array[...] = tensor_data
    return read_result.stamp

  return ts.virtual_chunked(
    read_function=read,  
    rank=len(chunk_info["chunks"]),
    shape=tuple(sum(dc) for dc in chunk_info["chunks"]),
    dtype=chunk_info["dtype"],
    chunk_layout=ts.ChunkLayout(chunk_shape=[c[0] for c in chunk_info["chunks"])
   )
```

Reading the above kvstore in a batch doesn't have the intended effect because the batch isn't propagated through VirtualChunked layer. It's probably impossible to guess how the batch will be used within the read functions (AFAICT, batches aren't used in write functions), but would it be possible to expose the batch on the VirtualChunked ReadParameters argument so that read functions can use the batch?

In the above example this would allow `read` to propagate the batch through the the `http_store.read` method via `params`.

sjperkins:
- Closes #276

### DIFF
--- a/tensorstore/driver/virtual_chunked/BUILD
+++ b/tensorstore/driver/virtual_chunked/BUILD
@@ -27,6 +27,7 @@ tensorstore_cc_library(
         "//tensorstore:open_mode",
         "//tensorstore:open_options",
         "//tensorstore:rank",
+        "//tensorstore:schema",
         "//tensorstore:staleness_bound",
         "//tensorstore:strided_layout",
         "//tensorstore:transaction",


python/tensorstore/tests/virtual_chunked_test.py
```diff
@@ -309,3 +309,37 @@ def do_read(
...
+      chunk_layout=ts.ChunkLayout(read_chunk_shape=(2, 3)),
+  )
+
+  import time
+
+  with ts.Batch() as b:
+    f = t[1:3, 1:3].read(batch=b)
+    # Immediate batch submission triggers
+    # asserts in Batch::Impl::~Impl()
+    time.sleep(.01)
```
sjperkins: Need to understand why this happens and whether it's possible to address this.

tensorstore/virtual_chunked.h
```diff
@@ -278,11 +279,15 @@ class ReadParameters {
...
 
+  /// Batch associated with read request
+  const std::optional<Batch>& batch() const { return batch_; }
+
   // Treat as private:
 
   Executor executor_;
   StorageGeneration if_not_equal_;
   absl::Time staleness_bound_;
+  std::optional<Batch> batch_;
```
sjperkins: The following also works

```suggestion
  Batch batch_;
```

but then the default constructor must be explicitly specified as 

```cpp
ReadParameters() : executor{}, ..., batch_{no_batch} {}
```

Not sure which is preferable.

tensorstore/virtual_chunked.h
```diff
@@ -278,11 +279,15 @@ class ReadParameters {
...
 
+  /// Batch associated with read request
+  Batch batch() const { return batch_; }
+
   // Treat as private:
 
   Executor executor_;
   StorageGeneration if_not_equal_;
   absl::Time staleness_bound_;
+  Batch batch_;
```
jbms: You can just specify the default value here.

laramiel:
> > There need to be additional tests in the C++ code for virtual_chunked/virtual_chunked_test.cc
> > I think that a few of the `Batch` parameters should instead be `Batch::View`.
> 
> Agreed on adding C++ tests. I don't know that I see anywhere that should use Batch::View --- since Batch::View behaves like string_view we can only use it where we are sure the batch will remain valid long enough.

I was thinking that `Batch batch() const { return batch_; }` could possible return `Batch::View`.

sjperkins:
> I was thinking that `Batch batch() const { return batch_; }` could possible return `Batch::View`.

This makes sense: I've implemented this change. After some thought I couldn't think of other cases, but I may have missed something.

tensorstore/driver/virtual_chunked/virtual_chunked.cc
```diff
@@ -237,7 +237,8 @@ void VirtualChunkedCache::DoRead(EntryOrNode& node,
   auto& executor = cache.executor();
   // `node` is guaranteed to remain valid until `ReadSuccess` or `ReadError`
   // is called.  Therefore we don't need to separately hold a reference.
-  executor([&node, staleness_bound = request.staleness_bound] {
+  executor([&node, staleness_bound = request.staleness_bound,
+            batch = Batch(request.batch)] {
```
laramiel: Needs an include of tensorstore/batch.h

tensorstore/driver/virtual_chunked/virtual_chunked_test.cc
```diff
@@ -587,4 +589,77 @@ TEST(VirtualChunkedTest, NonAtomicSingleChunk) {
...
+                                tensorstore::ReadWriteMode::read>>
+BatchSettingView(Batch& batch, Option&&... option) {
+  auto mutex = std::make_shared<absl::Mutex>();
+  return tensorstore::VirtualChunked(
+      tensorstore::NonSerializable{
+          [mutex, &batch](auto output, auto read_params)
+              -> Future<TimestampedStorageGeneration> {
+            tensorstore::InitializeArray(output);
+            absl::MutexLock lock(*mutex.get());
+            batch = read_params.batch();
```
laramiel: So batch is an output parameter? Name it as such.

laramiel: I'd start each of the three test cases like this to make the symmerty obvious

```
{
    Batch output_batch{Batch::no_batch};
    auto virtual_chunked =
         BatchSettingView(output_batch, tensorstore::dtype_v<int>,
                          tensorstore::Schema::Shape({2, 3}),
                          tensorstore::ChunkLayout::ReadChunkShape({2, 1}));
    ...
```

laramiel: So here are my reworked tests:

```
template <typename... Option>
Result<tensorstore::TensorStore<void, dynamic_rank,
                                tensorstore::ReadWriteMode::read>>
BatchSettingView(std::optional<Batch::View>& output_batch, Option&&... option) {
  auto mutex = std::make_shared<absl::Mutex>();
  return tensorstore::VirtualChunked(
      tensorstore::NonSerializable{
          [mutex, &output_batch](auto output, auto read_params)
              -> Future<TimestampedStorageGeneration> {
            tensorstore::InitializeArray(output);
            absl::MutexLock lock(*mutex.get());
            output_batch = read_params.batch();
            return TimestampedStorageGeneration{
                StorageGeneration::FromString("abc"), absl::Now()};
          }},
      std::forward<Option>(option)...);
}

// Test passing of Read batch into VirtualChunked ReadParameters
// No Batch argument used in tensorstore::Read
// results in no_batch set in virtual_batch
TEST(VirtualChunkedTest, ReadNoBatch) {
  std::optional<Batch::View> output_batch;
  auto virtual_chunked =
      BatchSettingView(output_batch, tensorstore::dtype_v<int>,
                       tensorstore::Schema::Shape({2, 3}),
                       tensorstore::ChunkLayout::ReadChunkShape({2, 1}));

  auto data = tensorstore::Read(virtual_chunked).result();
  EXPECT_TRUE(output_batch.has_value());
  EXPECT_FALSE(*output_batch);
}

// No Batch argument passed to tensorstore::Read
// results in no_batch set in virtual_batch
TEST(VirtualChunkedTest, ReadNoBatchArgument) {
  std::optional<Batch::View> output_batch;
  auto virtual_chunked =
      BatchSettingView(output_batch, tensorstore::dtype_v<int>,
                       tensorstore::Schema::Shape({2, 3}),
                       tensorstore::ChunkLayout::ReadChunkShape({2, 1}));

  auto batch = Batch{tensorstore::no_batch};
  auto read_future = tensorstore::Read(virtual_chunked, batch);
  EXPECT_FALSE(output_batch.has_value());
  batch.Release();
  auto data = read_future.result();
  EXPECT_TRUE(output_batch.has_value());
  EXPECT_FALSE(*output_batch);
}

// Batch argument passed to tensorstore::Read
// is reflected in virtual_batch
TEST(VirtualChunkedTest, ReadBatchArgument) {
  std::optional<Batch::View> output_batch;
  auto virtual_chunked =
      BatchSettingView(output_batch, tensorstore::dtype_v<int>,
                       tensorstore::Schema::Shape({2, 3}),
                       tensorstore::ChunkLayout::ReadChunkShape({2, 1}));

  auto batch = Batch::New();
  auto read_future = tensorstore::Read(virtual_chunked, batch);
  EXPECT_FALSE(output_batch.has_value());
  batch.Release();
  auto data = read_future.result();
  EXPECT_TRUE(output_batch.has_value());
  EXPECT_TRUE(*output_batch);
}
```

tensorstore/virtual_chunked.h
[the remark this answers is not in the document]
laramiel: TBH, my reverse mapping of build deps is somewhat broken right now, so I haven't worried about that; generally the bazel build cleaner tool should fix those if the includes are ~correct (within the same project), and that gets run on import.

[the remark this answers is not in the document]
sjperkins: > TBH, my reverse mapping of build deps is somewhat broken right now, so I haven't worried about that; generally the bazel build cleaner tool should fix those if the includes are ~correct (within the same project), and that gets run on import.

Still not understanding you I'm afraid. The bazel build cleaner seems to be an internal tool within Google?

If this works with the tensorstore build as is, can you please provide a command? i.e. ` python bazelisk.py run ...`

PR #287, opened 2026-03-16
Add documentation for open_as_void

BrianMichell:
Adds a formal definition for the `open_as_void` intended implementation behavior for #272 and #271


BrianMichell:
I've updated the definition to reflect the codec chain changes. Once we come to an acceptable definition here I will go back and update #271 implementation.

BrianMichell:
@jbms @laramiel Could I please get some feedback on this? Very eager to get the feature properly implemented and merged!

docs/open_as_void.rst
```diff
@@ -0,0 +1,245 @@
...
+
+For zarr v3, ``open_as_void`` operates entirely during the resolution of the
+codec pipeline. The implementation resolves the pipeline with a substituted
+"raw" data type and then validates the result, rather than checking each codec
+individually against an allowlist.
+
+1. **Data type becomes byte**: The array's data type is replaced with the
+   ``byte`` data type (a 1-byte, endian-invariant type) during codec pipeline
+   resolution, regardless of the original data type.
+2. **Additional dimension added**: A new innermost dimension is appended to
```
jbms: It should be noted that this additional dimension is added as a "field dimension" (or maybe it should be given a different name) that always remains the inner-most dimension even when there are array -> array codecs like transpose ---array -> array codecs only apply to the normal chunked dimensions, not field dimensions.

jbms:
One minor change: This open_as_void document could be moved to the zarr3/ directory rather than linked from the top-level doc index.   The zarr/index.rst can still reference it.  Other than that this documentation looks good.

PR #293, opened 2026-07-15
Require Python 3.11 in setup.py

### LINKED ISSUE #292: setup.py accepts Python 3.10 despite requiring Python 3.11+
## Description

The project metadata and installation documentation require Python 3.11 or newer:

- `pyproject.toml` specifies `requires-python = ">= 3.11"`.
- `docs/installation.rst` states that Python 3.11 or newer is required.

However, the explicit version check in `setup.py` only rejects versions older than Python 3.10 and reports:

```text
Python >= 3.10 is required to build
```

As a result, invoking `setup.py` directly with Python 3.10 proceeds past the version guard even though Python 3.10 is no longer supported.

This appears to be a stale check left after commit [`af2bd4f`](https://github.com/google/tensorstore/commit/af2bd4fd3aa58cfd442e92d3b0245581195eb79c) removed Python 3.10 support while adding Python 3.14 builds. That change updated `pyproject.toml`, the installation documentation, and wheel-building configuration, but not `setup.py`.

## Reproduction

From a TensorStore checkout, using Python 3.10:

```console
$ python3.10 setup.py --name
tensorstore
$ echo $?
0
```

## Expected behavior

`setup.py` should reject Python 3.10 and report that Python 3.11 or newer is required, consistently with `pyproject.toml` and the installation documentation.

## Actual behavior

Python 3.10 passes the explicit version check, and `setup.py` continues.

mattfaltyn:
## Summary

- Update the explicit `setup.py` version guard to require Python 3.11.
- Correct the corresponding error message.

## Root cause

Python 3.10 support was removed in [`af2bd4f`](https://github.com/google/tensorstore/commit/af2bd4fd3aa58cfd442e92d3b0245581195eb79c), but the direct `setup.py` guard was not updated alongside `pyproject.toml`, the installation documentation, and wheel-building configuration.

## Impact

Direct `setup.py` invocations now reject unsupported Python 3.10 consistently with the project metadata and documentation. Supported Python versions are unaffected.

## Validation

- Python 3.10: `setup.py --name` reports `Python >= 3.11 is required to build` and exits with status 1.
- Python 3.11: `setup.py --name` reports `tensorstore` and exits successfully.
- `git diff --check`
- Bazel: 458 of 459 targets passed in the full sandboxed run with `--macos_minimum_os=10.13`; the remaining `//python/tensorstore:fork_test` then passed both tests directly using `/private/tmp` after the sandbox encountered a macOS privacy denial on its default temporary directory.

Fixes #292

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,8 +18,8 @@ This invokes bazel via the included `bazelisk.py` wrapper script.
 
 import sys
 
-if sys.version_info < (3, 10):
-  print('Python >= 3.10 is required to build')
+if sys.version_info < (3, 11):
+  print('Python >= 3.11 is required to build')
   sys.exit(1)
 
 # Import setuptools before distutils because setuptools monkey patches


laramiel:
I saw your bug, but missed this pull request. This is essentially the same as https://github.com/google/tensorstore/commit/bc4fbc11f69ee5379b2f611a934663006e1f573f

